PR #1, opened 2016-05-19
Add projectile support

justbur:


tarsius:
It *might* be better to put these files into a `projectile/` subdirectory. What do you think?

justbur:
Fine with me. It might be nice to have a simple way to generate the directory if it doesn't exist. I just tried and projectile doesn't seem to create the directory automatically.

tarsius:
That seems like a good argument *not* to use subdirectories in most cases. Or do you see an alternative to create missing directories when the `var` and `etc` forms are evaluated? If not, then we would end up creating many directories, including directories for packages that many users don't install/use.

justbur:
I just threw something together in #2, but that's a good point about creating a bunch of empty directories. A simple rule might be enough here, like if there are more than x files use a directory.

tarsius:
Something else we might want to consider is to just submit patches against the packages whose files we would like to put into subdirectories. Essentially

```diff
+(make-directory (file-name-directory the-file) t)
```

That would have the advantage that users who don't use `no-littering` but still want to put some files into subdirectories would also benefit.

Of course those patches won't always be merged in a useful time frame (or at all). E.g. if the patch is against a built-in package, then we can't really wait until the next release in 1-2 years. So we should still implement a mechanism for creating the directories ourselves. #3 seems like a good idea.

With such a mechanism in place, we can start using a subdirectory whenever we see fit. So `x` should probably be `2`. But we shouldn't be dogmatic about it and decide on a case-by-case basis.

justbur:
That seems reasonable to me. I can't think of a reason the package shouldn't create the path for you or at least warn about it not existing. I think projectile was silently failing but I'd have to test it to know for sure. 

Closing this because I added projectile in a new commit in #3

PR #3, opened 2016-05-22
Create macro to replace existing implementation

justbur:
Creates environment for using helper functions var and etc as well as a
namespace for variables related to each third-party package. The third
party package files are put into corresponding subdirectories using
eval-after-load which will only create the directories if the
corresponding package is loaded.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -120,6 +120,7 @@ subdirectory). The paths are created using
 ;;; Built-in packages
 
 (without-littering nil
+  (setq abbrev-file-name              (var "abbrev_defs"))
   (setq auto-save-list-file-prefix    (var "saves-"))
   (setq backup-directory-alist        (list (cons "." (var "backups/"))))
   (setq bookmark-default-file         (var "bookmarks.el"))
--- a/no-littering.el
+++ b/no-littering.el
@@ -140,8 +140,10 @@ subdirectory). The paths are created using
 
 ;;; Third-party packages
 
-;; omit the package here because `smex-save-file' should be set before the
-;; package loads
+(without-littering projectile
+  (setq projectile-cache-file          (var "cache"))
+  (setq projectile-known-projects-file (var "bookmarks.eld")))
+
 (without-littering smex
   (setq smex-save-file (var "smex-items")))
 
--- a/no-littering.el
+++ b/no-littering.el
@@ -77,36 +77,73 @@ This variable has to be set before `no-littering' is loaded.")
   "The directory where packages place their persistent data files.
 This variable has to be set before `no-littering' is loaded.")
 
-(cl-flet ((etc (file) (expand-file-name (convert-standard-filename file)
-                                        no-littering-etc-directory))
-          (var (file) (expand-file-name (convert-standard-filename file)
-                                        no-littering-var-directory)))
-  (with-no-warnings ; many of these variables haven't been defined yet
+(defun no-littering-make-path (dir file)
+  "Returns path DIR/FILE, creating any non-existing
+directories (including parent ones) if necessary."
+  (unless (file-exists-p dir)
+    (make-directory dir t))
+  (expand-file-name (convert-standard-filename file) dir))
+
+(defmacro without-littering (package &rest body)
+  "Execute BODY in an environment with short helper functions,
+`var' and `etc', for creating no-littering paths specific to
+PACKAGE. If PACKAGE is non nil, BODY will only be executed if
+PACKAGE can be found in `load-path' using `locate-library'. Note
+this means that the load-path for all affected packages needs to
+be set up prior to using this macro for a given package.
+
+Within the environment, the forms \(var \"file\"\) and \(etc
+\"file\"\) each yield a path of the form \"$base/[package/]file\"
+where $base is either `no-littering-var-directory' or
+`no-littering-etc-directory' and package/ is a directory with the
+same name as the third-party package PACKAGE (use nil to omit the
+subdirectory). The paths are created using
+`no-littering-make-path'. For example,
+
+\(without-littering projectile
+  \(setq projectile-cache-file          \(var \"cache\"\)\)
+  \(setq projectile-known-projects-file \(var \"bookmarks.eld\"\)\)\)."
+  (declare (indent 1))
+  (let ((package-name (and package (symbol-name package))))
+    `(when (or (null ,package-name)
+               (locate-library ,package-name))
+       (cl-flet ((etc (apply-partially
+                       #'no-littering-make-path
+                       (concat
+                        no-littering-etc-directory ,package-name)))
+                 (var (apply-partially
+                       #'no-littering-make-path
+                       (concat
+                        no-littering-var-directory ,package-name))))
+         ,@body))))
 
 ;;; Built-in packages
 
-    (setq auto-save-list-file-prefix       (var "saves-"))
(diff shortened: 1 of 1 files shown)


tarsius:
I generally like the idea, but the devil is in the detail. So I'll start with some nitpicking.

no-littering.el
```diff
@@ -77,36 +77,64 @@ This variable has to be set before `no-littering' is loaded.")
   "The directory where packages place their persistent data files.
 This variable has to be set before `no-littering' is loaded.")
 
-(cl-flet ((etc (file) (expand-file-name (convert-standard-filename file)
-                                        no-littering-etc-directory))
-          (var (file) (expand-file-name (convert-standard-filename file)
-                                        no-littering-var-directory)))
-  (with-no-warnings ; many of these variables haven't been defined yet
-
-;;; Built-in packages
```
tarsius: I would like to keep the built-in and third-party packages separate and keep those headings. The built-in section should begin with a `no-littering-setq` form whose PACKAGE is `nil` and whose entries should be sorted alphabetically. After that come the package-specific `no-littering-setq`s, sorted alphabetically by PACKAGE.

no-littering.el
```diff
@@ -77,36 +77,64 @@ This variable has to be set before `no-littering' is loaded.")
...
-    (setq tramp-persistency-file-name      (var "tramp.el"))
-    (setq trash-directory                  (var "trash/"))
-    (setq url-configuration-directory      (var "url/"))
-
-;;; Third-party packages
-
-    (setq smex-save-file                   (var "smex-items"))
-    ))
+(defun no-littering-make-path (base-dir package file)
+  "Returns path BASE-DIR/[PACKAGE/]FILE, creating the necessary
```
tarsius: The `[PACKAGE/]` part doesn't seem correct. This always uses just `FILE`, but `FILE` may have the form `PACKAGE-DIRECTORY/FILE-NAME`. Generally I think the doc-strings still need some work.

no-littering.el
```diff
@@ -77,36 +77,64 @@ This variable has to be set before `no-littering' is loaded.")
...
+directories if necessary."
+  (let ((dir (concat base-dir (when package (symbol-name package)))))
+    (unless (file-exists-p dir)
+      (make-directory dir t))
+    (expand-file-name (convert-standard-filename file) dir)))
+
+(defmacro no-littering-setq (package &rest sym-val-pairs)
```
tarsius: Not sure about the name yet. While `setq` can be used to set multiple variables, Stefan recommends against doing it and that the ability to set multiple variables using a single form is only maintained for backward compatibility. Personally I have started to move away from setting multiple variables in one form (except in cases where a single expression is expected and where using multiple `setq`s would require the use of `progn`).

But I can't really think of a better name right now.

no-littering.el
```diff
@@ -77,36 +77,64 @@ This variable has to be set before `no-littering' is loaded.")
...
-    (setq url-configuration-directory      (var "url/"))
-
-;;; Third-party packages
-
-    (setq smex-save-file                   (var "smex-items"))
-    ))
+(defun no-littering-make-path (base-dir package file)
+  "Returns path BASE-DIR/[PACKAGE/]FILE, creating the necessary
+directories if necessary."
+  (let ((dir (concat base-dir (when package (symbol-name package)))))
```
tarsius: In such cases I use `and` instead of `when`, since this is about the value, not side-effects.

no-littering.el
```diff
@@ -77,36 +77,64 @@ This variable has to be set before `no-littering' is loaded.")
...
+  org-id-locations-file         (var "org/org-id-locations")
+  org-registry-file             (var "org/org-registry.el")
+  recentf-save-file             (var "recentf.el")
+  save-place-file               (var "saveplace.el")
+  savehist-file                 (var "savehist.el")
+  tramp-persistency-file-name   (var "tramp.el")
+  trash-directory               (var "trash/")
+  url-configuration-directory   (var "url/"))
+
+(no-littering-setq projectile
```
tarsius: Let's put that into a separate commit.

justbur:
I think maybe a macro could just be used to create and environment with short helper functions. The `setq`s could then be explicit in the body. 

There is one bigger issue though and that's that I don't think `eval-after-load` will always work because some package variables have to be set before loading. It would be nice if we could do this generally but I think to do `eval-before-load` we'd have to advise `require`

justbur:
@tarsius I tweaked the idea and tried to address your comments. The macro is just a wrapper to set up an environment now to use instead of assuming that you will be using `setq`.

tarsius:
What about calling the macro `without-litter`? ;-)

Sure this would violate the package-prefix-convention, but it would be cute. What do you think?

justbur:
I like it :-) but I thought `without-littering` was a little more natural so I picked that.

justbur:
Actually it would make sense to give the package the same name as this macro if you wanted to do that. Then we wouldn't be breaking convention

tarsius:
Sneaky:)

justbur:
@tarsius what do you think about the name change? Should we just leave everything as is?

tarsius:
I doubt that using a macro is the right approach in the first place. It seems like an unnecessary complication. The goal is to set all path variables in the file `no-littering.el`. We do not have to provide a "nice to use interface" for users to use in their own init file. We want them to contribute to this library instead.

You have pointed out that we have to guarantee that the package specific subdirectory exists if we want to instruct the package to save its files in there. Why don't we just do that?

Among the packages that have already been "theme", this only affects `org`. I think we should just tackle the issue directly like so:

```diff
diff --git a/no-littering.el b/no-littering.el
index b836b12..ad2e6d9 100644
--- a/no-littering.el
+++ b/no-littering.el
@@ -94,6 +94,8 @@ (cl-flet ((etc (file) (expand-file-name (convert-standard-filename file)
     (setq eshell-directory-name            (var "eshell/"))
     (setq ido-save-directory-list-file     (var "ido.last"))
     (setq nsm-settings-file                (var "network-security.data"))
+    (eval-after-load 'org
+      `(make-directory ,(var "org/") t))
     (setq org-id-locations-file            (var "org/org-id-locations"))
     (setq org-registry-file                (var "org/org-registry.el"))
     (setq recentf-save-file                (var "recentf.el"))
```

You can find that in the [explict](https://github.com/tarsius/no-littering/compare/explicit?expand=1) branch.

I have also attempted to write a macro which can be found in the [macro](https://github.com/tarsius/no-littering/compare/macro?expand=1) branch. But as I said, I think that using macro is unnecessary and complicates matters without actually providing any benefit.

justbur:
I didn't say it, but I thought that there might be a way to get package authors to support this naming convention in the packages themselves, in which case providing some sort of automated convention would be useful. Not sure if that means a macro is the right approach, but that was sort of my thought. 

In any case, I'm happy ditching the macro.

By the way, I don't think `eval-after-load` will be a good solution for no-littering. It's a problem for `projectile` and `smex` if the path doesn't exist on load, and it seems like this could be a common issue since it seems likely that loading a package would naturally invoke a routine to look for a history file.

justbur:
Just had another (simpler) idea. What if you just came up with a naming convention for the files with no subdirectories? Any file associated with PACKAGE would be prefixed with "PACKAGE-". For example, any file associated with projectile would need to start with "projectile-". That way there is some organization to the files (and you know which package each file is associated with) without worrying about managing the directories.

PR #4, opened 2016-05-26
Add support for projectile

sooheon:
Put projectile-cache-file and projectile-known-projects-file in var.


justbur:
Thanks, but I added this already as part of #3, which I think will be merged soon

justbur:
#3 is not going to be merged, so if you want to reopen this it please do.

PR #5, opened 2016-05-27
Name files based on variable names

justbur:
This will give the files prefixes that correspond to the packages they
relate too. It will also be easier to track down the relevant variable
when all you have is the file name.


justbur:
@tarsius just another kind of approach to consider

tarsius:
Sorry for all the delays. I have my head somewhere else right now, but will try to get back to this soon, possibly today.

no-littering.el
```diff
@@ -85,27 +85,27 @@ This variable has to be set before `no-littering' is loaded.")
...
-    (setq recentf-save-file                (var "recentf.el"))
-    (setq save-place-file                  (var "saveplace.el"))
+    (setq recentf-save-file                (var "recentf-save.el"))
+    (setq save-place-file                  (var "save-place.el"))
     (setq savehist-file                    (var "savehist.el"))
-    (setq tramp-persistency-file-name      (var "tramp.el"))
+    (setq tramp-persistency-file-name      (var "tramp-persistency.el"))
     (setq trash-directory                  (var "trash/"))
-    (setq url-configuration-directory      (var "url/"))
+    (setq url-configuration-directory      (var "url-configuration/"))
```
tarsius: > This will give the files prefixes that correspond to the packages they relate too.

Okay let's do that for regular files, but not for directories. Here we should stick to the package name.

justbur:
Closed in favor of #6

PR #6, opened 2016-05-31
Another attempt

tarsius:


justbur:
:+1:

tarsius:
I'm not 100% convinced of the idea to base file names on the variable name. If a package has just one file (or two one in var, one in etc), then we could just stick to the package/library name.

tarsius:
I mean it wouldn't be to hard to determine the appropriate variable even if we did the latter. Are there other advantages to the former approach?

justbur:
If I see a random file that got created in var for instance it would be
helpful if the name of the file indicated the variable name I should search
for in emacs. Doesn't have to be 1:1 but a pointer in the right direction
would be nice. I figured that the actual names of the files are not that
important so we might as well make them as useful as possible. At the very
least they and the directories should be prefixed with the library name I
think.

justbur:
Take smex for example. What's the advantage to naming it smex-items when smex-save more closely matches the variable name? If I want to know what the file is I should be looking at the docs for the variable in emacs anyway, right?

PR #7, opened 2016-05-31
Add Conventions to README

justbur:
@tarsius This is what I was thinking for naming conventions, but feel free to change/criticize/ignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,3 +40,25 @@ variables before loading the feature.
     (setq no-littering-var-directory
           (expand-file-name "data/" user-emacs-directory))
     (require 'no-littering)
+
+Conventions
+-----------
+
+### Files
+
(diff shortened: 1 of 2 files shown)


tarsius:
I've made some minor adjustments mostly regarding punctuation.

The biggest change is that I dropped "/feature" from "Any subdirectories at the top-level should use the corresponding package/feature name as a prefix, similar to how files are named.". I think that we usually should use just the package name, e.g. if the package `foo` implements the code responsible for saving stuff in files is implemented in `foo-persistence.el`, then that is an implementation detail (which may change) and we should just put the files into `foo/` instead of `foo-persistence`.

If also copied the documentation into the library header and adjusted the commit message accordingly.

I hope you are okay with me having made these changes to your commit.

PR #8, opened 2016-06-02
Add theme for anaconda and undo-tree

sooheon:
Anaconda server installation directory and undo-tree history folder.


tarsius:
Thanks. I have split the commit into two, one for each package. Please do so too for future additions, and keep the lists of variables sorted alphabetically.

PR #10, opened 2016-06-05
Theme various packages

tarsius:
This themes the same packages as #9 but it follows the conventions more strictly and themes all of the relevant variables of the themed packages, not just a subset.

Please see the commit messages for justifications for the decisions I made in cases where the conventions aren't all that clear yet.


tarsius:
@justbur Should we always use `<package>/<rest>` instead of `<package>/<package>-<rest>` even if `<package>` is as short as `org`?

justbur:
My preference would be to stick to a standard and not introduce exceptions unless they are necessary.

tarsius:
So you are in favor of dropping the `org-`?

tarsius:
Hm, actually that wouldn't prevent inconsistency because some packages don't allow customizing the file names using a "directory"-variable. So we would have to have always use the prefixes to be consistent. I guess that's better, but a bit noisy in some cases.

justbur:
As long as there are some rules that are clear I don't care too much about how they are chosen. If we have clear rules then it makes PRs very simple to merge without much review.

justbur:
By the way I'm sure you've run into this but what should we say about migration to this package since it effectively creates a bunch of new history files where the old ones might be pretty valuable to people?

justbur:
I'd say the directory should count as the prefix, but at the same time we have little control over what goes into the directory sometimes so we cannot be strict about it. My preference would be to treat the directory as the prefix and not "double prefix" the files in subdirectories if we are picking the names for them.

tarsius:
> I'd say the directory should count as the prefix,

Good, that's my preference too. I have pushed a commit which changes this for Org and will update the documentation tomorrow.

> what should we say about migration to this package since it effectively creates a bunch of new history files where the old ones might be pretty valuable to people?

I don't think we can help here much at least not in an automated fashion, that would lead right into complicated-macro-land again. So I guess we should warn users that they have to move files manually or (and that's what I would recommend because I think in most cases the data isn't actually *that* valuable) delete the old files.

That's a bit unfortunate, but I don't see a simple way to avoid it.

PR #12, opened 2016-06-11
Patch 1

xuchunyang:


xuchunyang:
BTW, notice that the Emacs built-in tutorial.el uses a function to decide where to keep its cache.
```el
(defun tutorial--saved-dir ()
  "Directory to which tutorials are saved."
  (locate-user-emacs-file "tutorial/"))
```

tarsius:
I've pushed commits which theme these packages to `master`. In most cases I made some changes to better comply with the conventions (including some conventions that haven't been formalized yet).

PR #13, opened 2016-06-14
Add persistent-scratch

benjaminfrank:


benjaminfrank:
persistent-scratch also has `persistent-scratch-backup-directory` but semantics is to keep backups once the folder is non-nil. I think setting this as default might not what a user expects and thus omited it in this PR.

justbur:
Thanks, but we're trying to keep the variables for third-party packages in alphabetical order.

justbur:
> persistent-scratch also has persistent-scratch-backup-directory but semantics is to keep backups once the folder is non-nil. I think setting this as default might not what a user expects and thus omited it in this PR.

Yes, that's not something we would want to choose for them I don't think.

benjaminfrank:
Changed order.

justbur:
Looks good. Thanks

PR #14, opened 2016-06-14
Add ensime

benjaminfrank:
Not sure if you want the order to be project `ensime` or var `user-emacs-ensime-directory`


justbur:
Not sure either ;). Let's see what @tarsius thinks.

tarsius:
Order by variable name for now (we can still change it later if we have to add other ensime variables).

benjaminfrank:
Changed order accordingly.

justbur:
Thanks!

PR #15, opened 2016-06-14
company-statistics: theme company-statistics-file

swarnendubiswas:
Added a configuration for company-statistics-file.


tarsius:
We keep these lists sorted alphabetically, so please move your addition to the correct place. Also please use `(var "company/statistics.el")`.

swarnendubiswas:
I will do that, but shouldn't we name the file `company/statistics-cache.el`?

justbur:
Thanks!

justbur:
In most cases I think "cache" is analogous to "file", so to fit the conventions it makes sense to drop it.

PR #23, opened 2016-11-16
Fix desktop, add eww and irony

xuchunyang:


tarsius:
Thanks!

I have changed the order in the second commit to match similar existing parts.

PR #25, opened 2016-11-27
Add directories for emojify and xkcd

therockmandolinist:


tarsius:
Thanks! I have spread the fixup commit across the preceding commits before merge.

PR #26, opened 2016-11-30
auto-save: Move files into a directory

justbur:
auto-save files get created for each new PID and this creates quite a bit of
clutter in var.

What do you think @tarsius? I did this because my var directory was filling up with these files.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -70,7 +70,7 @@
 ;;
 ;; 1. File names are based on the name of the respective Emacs lisp
 ;;    variables and the name of the respective Emacs package.
-;;   
+;;
 ;; 2. The name of the respective Emacs package should serve as the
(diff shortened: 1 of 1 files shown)


therockmandolinist:
In my emacs (26.0.50.1) auto save files get automatically placed in the temp directory, with only the list of auto saves being placed in var. I think this is the only thing enforced in no-littering, so curious as to how your auto saves are ending up in var.

justbur:
It's the list files I'm referring to here.

therockmandolinist:
Oh, I see. How come there's more than one? I don't have too much experience with the auto save list files themselves, but I think so far I've had one at a time for the most part.

justbur:
They are unique by PID, so if you start up multiple instances of emacs you get more than one.

PR #27, opened 2016-11-30
Add bbdb, elfeed, and elfeed-org

therockmandolinist:
I've grouped the last two under the "elfeed" folder since elfeed-org is an enhancement to elfeed.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -234,6 +234,7 @@ This variable has to be set before `no-littering' is loaded.")
        `(make-directory ,(var "elfeed/") t))
     (setq elfeed-db-directory              (var "elfeed/db/"))
     (setq elfeed-enclosure-default-dir     (var "elfeed/enclosures/"))
+    (setq rmh-elfeed-org-files             (list (var "elfeed/elfeed.org")))
     (setq emms-directory                   (var "emms/"))
     (eval-after-load 'emojify
       `(make-directory ,(var "emojify/") t))
--- a/no-littering.el
+++ b/no-littering.el
@@ -230,6 +230,10 @@ This variable has to be set before `no-littering' is loaded.")
     (eval-after-load 'company-statistics
       `(make-directory ,(var "company/") t))
     (setq company-statistics-file          (var "company/statistics.el"))
+    (eval-after-load 'elfeed
+       `(make-directory ,(var "elfeed/") t))
+    (setq elfeed-db-directory              (var "elfeed/db/"))
+    (setq elfeed-enclosure-default-dir     (var "elfeed/enclosures/"))
     (setq emms-directory                   (var "emms/"))
     (eval-after-load 'emojify
       `(make-directory ,(var "emojify/") t))
--- a/no-littering.el
+++ b/no-littering.el
@@ -218,6 +218,10 @@ This variable has to be set before `no-littering' is loaded.")
 
     (setq ac-comphist-file                 (var "ac-comphist.el"))
     (setq anaconda-mode-installation-directory (etc "anaconda-mode/"))
+    (eval-after-load 'bbdb
+      `(make-directory ,(var "bbdb/") t))
+    (setq bbdb-file                        (var "bbdb/bbdb.el"))
+    (setq bbdb-vcard-directory             (var "bbdb/bbdb-vcard/"))
     (setq bm-repository-file               (var "bm-repository.el"))
     (eval-after-load 'bookmark+-1
       `(make-directory ,(var "bmkp/") t))


tarsius:
> tarsius approved these changes just now

No, I approved one out of three commits.

no-littering.el
```diff
@@ -217,6 +217,10 @@ This variable has to be set before `no-littering' is loaded.")
 
     (setq ac-comphist-file                 (var "ac-comphist.el"))
     (setq anaconda-mode-installation-directory (etc "anaconda-mode/"))
+    (eval-after-load 'bbdb
```
tarsius: *meta*

(1) A line comment on a single commit, using "Start a review".

no-littering.el
```diff
@@ -217,6 +217,10 @@ This variable has to be set before `no-littering' is loaded.")
...
+      `(make-directory ,(var "bbdb/") t))
+    (setq bbdb-file                        (var "bbdb/bbdb.el"))
+    (setq bbdb-vcard-directory             (var "bbdb/bbdb-vcard/"))
```
tarsius: *meta*

(2) And another, using "Add review comment".

tarsius:
*meta*

(4) Ah, that's actually useful. Let's experiment some more.

Except that I cannot comment on the review, only individual line comments or here at the top-level.

no-littering.el
tarsius: *meta*

(5) Reply to a line-comment.

no-littering.el
```diff
@@ -217,13 +217,22 @@ This variable has to be set before `no-littering' is loaded.")
...
     (eval-after-load 'bookmark+-1
       `(make-directory ,(var "bmkp/") t))
     (setq bmkp-current-bookmark-file       (var "bmkp/current-bookmark.el"))
     (setq bmkp-last-bookmark-file          (var "bmkp/last-bookmark.el"))
     (eval-after-load 'company-statistics
       `(make-directory ,(var "company/") t))
     (setq company-statistics-file          (var "company/statistics.el"))
+    (eval-after-load 'elfeed
```
tarsius: *meta*

(6) Line-comment on "Files changed" and then "Start a review".

tarsius:
*meta* (I am stopping that now, and will open a new pr for the actual changes later)

(8) The 6,7 review takes me to the overview and the 1,2,3 review takes me to the individual commit. Good.

tarsius:
(9) What happens if I close...

tarsius:
(10) ... and reopen?

tarsius:
(11) Review still there. Good.

tarsius:
(12) Going to force push, rebased but unmodified commits...

tarsius:
(13) Actually let's make a non-review comment first...

no-littering.el
```diff
@@ -217,13 +217,22 @@ This variable has to be set before `no-littering' is loaded.")
 
     (setq ac-comphist-file                 (var "ac-comphist.el"))
     (setq anaconda-mode-installation-directory (etc "anaconda-mode/"))
+    (eval-after-load 'bbdb
+      `(make-directory ,(var "bbdb/") t))
+    (setq bbdb-file                        (var "bbdb/bbdb.el"))
```
tarsius: (14) Non-review line comment.

tarsius:
(14) Pushing now...

tarsius:
(15) ... or not, it seems you have disabled this.

@therockmandolinist Please rebase and then push, so that I can continue this experiment.

tarsius:
(16) Old commits are still thrown away, but at least the diffs are kept. Sometimes that's enough...

... sometimes not. Let's see if the old mechanism for preventing that still works:

"Mentioning the current tip: 4e22b51f081381dfbc9577a066687338e3cf2af7."

no-littering.el
```diff
@@ -234,6 +234,7 @@ This variable has to be set before `no-littering' is loaded.")
        `(make-directory ,(var "elfeed/") t))
     (setq elfeed-db-directory              (var "elfeed/db/"))
     (setq elfeed-enclosure-default-dir     (var "elfeed/enclosures/"))
+    (setq rmh-elfeed-org-files             (list (var "elfeed/elfeed.org")))
```
tarsius: (17) Will this be preserved?

tarsius:
(19) No.

no-littering.el
tarsius: (20) But somehow this comment got moved from 4e22b51f081381dfbc9577a066687338e3cf2af7 to dd162629c6bc3d06bd8dca9cf800a75d58475c08, which is at least surprising.

tarsius: (21) What if I mention its hash when commenting on that commit? That would be dd162629c6bc3d06bd8dca9cf800a75d58475c08 at this time. Using "Add single comment".

tarsius:
(22) Pushing...

tarsius:
(23) Still no. Well at least the diffs are there and if we really want to be able to get our hands on the old tip, then we can still mention its hash to preserve it. However we have lost the following:

1. From the comment it is not possible to extend the diff to see the context.
2. From the commit it is not possible to see the comment.

But I guess we can open two windows.

therockmandolinist:
Sorry, could you clarify which part you don't like? Not sure I understand. Is it simply the repetition of `bbdb` in the file name and folder? They could be renamed.

tarsius:
> Sorry, could you clarify which part you don't like? Not sure I understand. Is it simply the repetition of `bbdb` in the file name and folder? They could be renamed.

Yes.

> [File names (2)] The name of the respective Emacs package should serve as the prefix of the file name, unless the file is in a subdirectory in which case the name of the subdirectory serves as the prefix.

Implies that the prefix should not be repeated. But if we remove that, what remains of `bbdb-file` if we also follow

> [File names (4)]  If the name of a path variable ends with -file, -default-file, -directory, -default-directory, or something similar, then that suffix is usually dropped from the file name.

`-`?!

Normally that would not be an issue because of

> [File location and subdirectories (5)] If a package only defines a single variable that specifies a data (or config) directory, then the directory name should nevertheless be just the package name.

But because of

> [File location and subdirectories (4)] A package that provides a "framework" for other packages to use, then we may reuse its directories for other packages that make use of that framework or otherwise "extend" the "main package". 

it is in this case.

tarsius:
wip

tarsius:
* There's no way around `bbdb/bbdb.el`, but I changed `bbdb/vcard.el` instead of `bbdb/bbdb-vcard.el`.
* I also moved the definition of `rmh-elfeed-org-files` and changed the path to`elfeed/rmh-elfeed.org`.

PR #28, opened 2016-12-09
Add auto-insert and bm

glucas:


tarsius:
Thanks! I have cherry-picked those commits.

PR #29, opened 2016-12-12
Move url-configuration-directory to var

glucas:
I think `url-configuration-directory` should be set to a `var` directory, rather than `etc`. Despite having "configuration" in the name, what gets stored there is things like cookies and history. That seems more like cache state than configuration.

I propose using var/url and var/url/cache, but perhaps it would be more consistent with the no-littering guidelines to have var/url/configuration and var/url/cache.


tarsius:
> perhaps it would be more consistent with the no-littering guidelines to have var/url/configuration and var/url/cache.

I've changed your commit to do that and applied that.

Thanks!

PR #30, opened 2016-12-31
speed-type: theme speed-type-gb-dir

therockmandolinist:


tarsius:
Do we really have to create this directory? I would expect that the package takes care of that itself.

therockmandolinist:
I'm actually unsure, here's the commit that added this functionality: parkouss/speed-type@f87399c

tarsius:
Looking at the rest of the code in `speed-type--gb-retrieve` revealed that it is not. I have adjusted your commit and pushed to `master`.

PR #32, opened 2017-01-17
jabber: theme jabber-avatar-cache-directory and jabber-history-dir

DamienCassou:


tarsius:
Thanks! I have cherry-picked.

PR #33, opened 2017-01-17
runner: theme runner-init-file

DamienCassou:


tarsius:
It doesn't seem likely that this package gains a second data file, so I am using `runner-init.el`, no sub-directory.

PR #34, opened 2017-01-17
org-caldav: theme variables

DamienCassou:
Theme `org-caldav-backup-file` and `org-caldav-save-directory`.


tarsius:
I have rearranged the lines.

PR #37, opened 2017-01-22
Correct a typographical error in the package conventions

doolio:
Correction applied to both the Emacs Lisp and Markdown files. Removed some
trivial errors highlighted by Flycheck as well.


tarsius:
Thanks!

PR #41, opened 2017-02-16
save-visited-files: theme save-visited-files-location

jpkotta:


tarsius:
I have changed the file name to be equal to the file name. Which is a bit weird, but that's because the variable has a weird name. I didn't like how the name you picked ended with `-files` as that implies that it is a directory containing multiple files.

PR #42, opened 2017-02-26
persp-mode: theme persp-save-dir

ellakk:


tarsius:
I have dropped the `make-directory` as the package takes care of that itself.

I think the file name is a bit weird, but it matches our conventions, so I am sticking with that.

PR #43, opened 2017-03-06
auto-save: theme auto-save-list-file-prefix and auto-save-file-name-t…

jpkotta:
This puts autosave files in var/auto-save/ and autosave lists in var/auto-save-lists.  Previously, autosave lists went in var/auto-save and autosave files went in the PWD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,17 @@ the following.
     (add-to-list 'recentf-exclude no-littering-var-directory)
     (add-to-list 'recentf-exclude no-littering-etc-directory)
 
+#### Auto-save settings
+
+One of the most common types of files that Emacs creates automatically
+is auto-save files.  By default, these appear in the current directory
+of a visited file.  No-littering does not change this, but you can add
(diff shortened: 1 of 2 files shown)


tarsius:
I am not so sure we should do this. Placing an autosave file in the same directory as the real file has the advantage that the user can see that there are edits that haven't been saved yet. Some users might want to stick to that.

We might want to add something like

```lisp
(defun no-littering-var-directory (file &optional create-directory)
  (setq file (expand-file-name (convert-standard-filename file)
                               no-littering-var-directory))
  (when create-directory
    (make-directory (file-name-directory file) t)))
```

and suggest that users who want what you are suggesting do this:

```lisp
(setq auto-save-file-name-transforms
      (list (list ".*" (no-littering-var-directory "auto-save/") t)))
```

jpkotta:
ping

tarsius:
Sorry for the delay. I wanted to do things a bit differently but didn't get around to it until now. See #46.

PR #44, opened 2017-03-06
tramp: theme tramp-auto-save-directory and tramp-backup-directory-alist

jpkotta:
tramp-auto-save-directory puts autosaves for remote files on the local machine
tramp-backup-directory-alist is like backup-directory-alist, but applies to remote files.  These files are actually backed up on the remote system, roughly (concat tramp-prefix matching-dir-from-alist backup-name).


tarsius:
`s:"tramp-:"tramp/` as per "File location and subdirectories" 2 and 3.

jpkotta:
ping

tarsius:
Sorry for the delay. I wanted to do things a bit differently but didn't get around to it until now. See #46.

PR #45, opened 2017-03-09
historian: theme `historian-save-file'

raxod502:
Excellent new package for replacing `smex`.


tarsius:
Could you please first add it to Melpa? Also I could not find the upstream repository, could you please point me to it?

tarsius:
Applied. Thanks!

PR #48, opened 2017-05-29
tldr: theme tldr-directory-path

therockmandolinist:


tarsius:
```lisp
(defvar tldr-pages-dir (concat tldr-directory-path "pages/")
  "Don't change me you idiot!")
```

Despite the insult, this variable actually has to be set. Also it should be defined using `defcustom` and the definition should contain `:set-after '(tldr-directory-path)`. Or better yet, this variable should be removed completely. Please discuss this with upstream.

tarsius:
[tldr](https://github.com/kuanyui/tldr.el)

/cc @kuanyui

kuanyui:
Sorry for the insult, I did never assume that docstring will be read. 😅 
However, I don't understand what this project exactly is doing....@@

therockmandolinist:
@tarsius I'm not sure I understand - won't the pages dir just be placed wherever tldr-directory-path? And what would be the advantage of removing it (not that I see why it's necessary either, just wondering)?

tarsius:
Applied because the mentioned issue does not actually affect `no-littering` because `no-littering` should be loaded before the packages it modifies, so https://github.com/kuanyui/tldr.el/pull/21 has no effect if users do things as instructed.

But that issue should still be fixed for non-no-littering users.

PR #49, opened 2017-06-15
indium: theme indium-workspace-file

NicolasPetton:


no-littering.el
```diff
@@ -262,6 +262,9 @@ directories."
     (setq helm-adaptive-history-file       (var "helm/adaptive-history.el"))
     (setq helm-github-stars-cache-file     (var "helm/github-stars-cache.el"))
     (setq historian-save-file              (var "historian-save.el"))
+    (eval-after-load 'indium
+      `(make-directory ,(var "indium/") t))
+    (setq indium-workspace-file (var "indium/indium-workspaces.el"))
```
DamienCassou: Do you need a directory if it's for one file only? If you want the directory, should the file be named "workspace.el" instead of "indium-worspace.el"?

NicolasPetton:
Damien Cassou <notifications@github.com> writes:

> Do you need a directory if it's for one file only? If you want the
> directory, should the file be named "workspace.el" instead of
> "indium-worspace.el"?

Indium will most probably set other files in the near future.

Regarding the name, I don't know.

no-littering.el
tarsius: Please teach `indium` to create the directory itself.

> should the file be named "workspaces.el" instead of "indium-worspaces.el"?

The former. (Assuming it contains information about multiple "workspaces". Otherwise just `workspace.el`.)

NicolasPetton: Fixed in https://github.com/nicolaspetton/indium/commit/3728af9

PR #53, opened 2017-06-25
org: theme org-publish-timestamp-directory

geoffmacintosh:
I hope I've followed the conventions correctly.

[Org-mode link](http://orgmode.org)

As per Emacs' documentation, this variable is set to `~/.org-timestamps/` which is kind of a weird place. I've had it changed in my config for months now, with no obvious issues, so I hope this is good to go. Thanks!


tarsius:
Thanks!

Applied with a slash appended to the filename.

PR #54, opened 2017-07-27
vimish-fold:theme vimish-fold directory

deadite81:
Add Vimish Folds persistent cache dir.  VF creates dir automatically.

Consequently fix errant whitespace.

https://github.com/mrkkrp/vimish-fold


tarsius:
Thanks!

I have done the whitespace cleanup in a separate commit and cleaned up your commit message.

PR #56, opened 2017-08-23
Theme lookup, async-bytecomp, kkc, notmuch, geiser, quack

stepnem:


tarsius:
I have adjusted the paths to follow the conventions. So far I have only taken the name of the variables into account. I will have to additionally check whether the files contain s-expressions (in which case the file names should end with `.el`) or not, and whether there are other config/data variables (in which case the file should be placed in a subdirectory, and other variables should likely also be themed).

I won't have time to work on this for at least a week. Including reviewing adjustments that you make, though that would be much appreciated never-the-less. If you do investigate, then please report your findings in the commit messages, e.g. "This is the only configuration variable of this package." and "The content of the file is a single s-extression.".

stepnem:
Thank you for the response.

I wouldn't think of making a pull request without first reading
(repeatedly) the conventions and checking the points you mention. (And
running Emacs with the modifications for a while.) Which is not to say
I understood them or didn't make any mistakes, of course. In any case,
this exchange will hopefully clarify some points for the future, see
below.

A general note: instead of making a modified commit with my name, I'd
much prefer either having committed the original unchanged and your
modification on top of it (ideally with explanation in the commit
message), or just committing whatever you want with only your name.
IMO adding "Modified-by" is of little use, at least without further
explanation.

Notes on your modifications follow. In general, it seems obvious that
I try not to change the original beyond necessity (following the
no-littering conventions), while you seem to prefer duplicating as
much as possible of the variable name in the file name (which, if a
general requirement/preference, should be mentioned in the
no-littering commentary section).

Please add the additional requirements (such as the particular details
to be reported in the commit message) you mention in your comments to
this PR to the commentary section of no-littering.el to save resources
(yours as well as those of potential contributors).

(I see now you have partially addressed this in c1e2cd9. Please also
note that not everyone who submits a PR is using the GitHub web
interface, so information only in the PR template is less likely to
be noticed than the library commentary.)

* geiser https://github.com/emacscollective/no-littering/commit/e9486e87d3e7dda7f688e96afd06ccb4bf376fa2
ACK

library original: (expand-file-name "~/.geiser_history")

* notmuch https://github.com/emacscollective/no-littering/commit/155c5c573ef3c7a11706a4e5c52d88c5238aef36
NAK

I dislike this for several reasons:

There is only one notmuch config file AFAICT, so according to the
conventions it should be a single file under etc.

The original commit message makes no sense any more (.el).

However intellectually pleasing it may be WRT the conventions, I don't
like many files with the same base name. It's a pain e.g. when using
virtual (i.e. now closed buffers of previously visited files) buffers
completion ("Oh noez, this is the notmuch init.el, not my Emacs init
file"). But there's nothing to be done I suppose.

library original: (locate-user-emacs-file "notmuch-config")

* kkc https://github.com/emacscollective/no-littering/commit/e55f1493fc769c5ec20af59fa24c201d8bb7c05a
ACK

library original: (locate-user-emacs-file "kkcrc" ".kkcrc")

* async-bytecomp https://github.com/emacscollective/no-littering/commit/b1b00176933cf4bc5039548b14906c03b8854a49
NAK

This is wrong as per the conventions (File names, section 3); the
library name is async-bytecomp, despite the misleading variable name.

library original: "~/.emacs.d/async-bytecomp.log"

tarsius:
> A general note: instead of making a modified commit with my name, I'd
> much prefer either having committed the original unchanged and your
> modification on top of it

I don't knowingly add changes to the mainline that I know will have to be changed.

> or just committing whatever you want with only your name.

I didn't do that because I figured most contributors would appreciate some attribution. However it would probably be better to create a new commit under my name and add a `Suggested-by: you` line as a form of attribution.

For commits such as the one where you edited `borg.org` but without regenerating `borg.texi`, I will continue to go with the modified-by approach.

tarsius:
- geiser: ack ack
- notmuch: use `notmuch-init`
- kkc: ack ack
- async-bytecomp

  Well, I would say our file name conventions do not account for variables that don't follow the variable name conventions :wink: I am undecided what to do here.

stepnem:
> - async-bytecomp
>
>   Well, I would say our file name conventions do not account for
>   variables that don't follow the variable name conventions :wink: I
>   am undecided what to do here.

To me this seemed a no-brainer -- there are many elisp libraries who fail
to stick to a single prefix in defined identifiers*, or use a prefix
different from the name of the library, but the conventions are quite
clear on this point: the file name prefix should be the same as the name
of the library.

* in the case at hand, the library async-bytecomp uses both
  async-bytecomp- and async-byte-, though the former is in majority. I
  suppose one could argue that the fact that async-bytecomp.el is
  included in the async package gives it the privilege to use the bare
  async- prefix, but I don't see why that should prevent us from keeping
  with the conventions. From another POV, I see a value of this
  particular convention in being able to determine the library that
  created a particular file just from the file name itself.

stepnem:
On Thu, 24 Aug 2017 03:08:11 -0700
Jonas Bernoulli wrote:

> I didn't do that because I figured most contributors would appreciate
> some attribution. However it would probably be better to create a new
> commit under my name and add a `Suggested-by: you` line as a form of
> attribution.

Sounds about good, thank you, though I really wouldn't mind no
attribution at all in such a case.

> For commits such as the one where you edited `borg.org` but without
> regenerating `borg.texi`, I will continue to go with the modified-by
> approach.

Yeah, I'm sorry, I just haven't invested the resources yet to find out
what exactly I have to do to regenerate it. Whether I have to install
the somewhat mysterious ox-texinfo+ first, for example.

tarsius:
These variables are finally themed on `master`.

I kept some of your commits unchanged (except that the were cherry-picked, resulting in different hashes). The commit that I have modified/replaced are attributed to me and only me, as you requested.

The `notmuch` commit contains a rant (though it also serves as documentation).

Speaking of rants: sorry for my initial response - you actually looked into complications more than most contributors before you have and my own initial changes were *less* well researched than your initial versions (though in my defense I stated that upfront).

Anyway, thanks and cheers!

stepnem:
Thanks for all the work!

The fact that notmuch was doing something wrong or at least very strange
was obvious to me as well, but same as when changing the file names, I
was tending to the path of least resistance (follow the conventions
while preserving as much of the original as possible). I suppose you
reported your findings upstream as well?

tarsius:
Last time I checked the upstream repository no longer existed and I am rather busy with the Magit kickstarter campaign right now, so no.

PR #58, opened 2017-09-14
Remove extra paren from auto-save settings section

YourFin:
There appeared to be an extra paren, which may cause errors for serial copy pasters like me.


tarsius:
Thanks!

I have applied this after improving the commit message. (It should not contain wording like "this probably does the right thing", and because you created the commit on Github, it lacked a final newline.)

PR #60, opened 2017-10-16
omnisharp: theme omnisharp-cache-directory

DamienCassou:
I think I followed all conventions. Please tell me if I didn't. The only thing I'm not sure about is if the subdirectory "cache/" is necessary or if "omnisharp/" would have been enough. There is no other variable to theme.

> please link to the repository of the package that your pull request is theming.

- https://github.com/OmniSharp/omnisharp-emacs

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -226,6 +226,7 @@ This variable has to be set before `no-littering' is loaded.")
     (setq nsm-settings-file                (var "nsm-settings.el"))
     (eval-after-load 'org
       `(make-directory ,(var "org/") t))
+    (setq omnisharp-cache-directory        (var "omnisharp/cache"))
     (setq org-clock-persist-file           (var "org/clock-persist.el"))
     (setq org-id-locations-file            (var "org/id-locations.el"))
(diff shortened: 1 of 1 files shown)


tarsius:
I cannot find that symbol anywhere in that package - not even using `rgrep` or `git log -G omnisharp-cache-directory`. I sense there must be a typo or something, but I don't see it :wink:

DamienCassou:
It works for me: https://github.com/OmniSharp/omnisharp-emacs/search?utf8=%E2%9C%93&q=omnisharp-cache-directory&type=.

tarsius:
Ah! Didn't update the Emacsmirror (or more specifically my local copy) in the last ~~24~~ 23 hours. :stuck_out_tongue_closed_eyes:

tarsius:
> The only thing I'm not sure about is if the subdirectory "cache/" is necessary or if "omnisharp/" would have been enough. There is no other variable to theme.

Considering how recently this one was added and how large the package is, it is conceivable that more path variables will be added. So it seems safest to do what you have done.

PR #63, opened 2017-11-28
litable: theme litable-list-file

dakra:
- This file keeps track of which functions (beyond the defaults) you trust to be pure.
- This is the only configuration/data file of the package


tarsius:
Thanks! Rebased.

PR #64, opened 2017-11-28
helm-backup: theme helm-backup-path

dakra:
Location where all the folders for the backup git repositories are.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -283,6 +283,7 @@ directories."
     (eval-after-load 'helm
       `(make-directory ,(var "helm/") t))
     (setq helm-adaptive-history-file       (var "helm/adaptive-history.el"))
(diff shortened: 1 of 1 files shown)


tarsius:
I have appended a slash at the end of the path.

Could you please fix this upstream:

1. The default value does not end with a slash.
2. But one place it does: `(concat helm-backup-path (file-name-directory filename))`.
3. One should always use `expand-file-name`, which is cross-platform and prevents bugs like (2).

tarsius:
And there is other path related weirdness too, e.g. `helm-backup--is-absolute-filename`.

dakra:
Thanks for the feedback. I made a PR here https://github.com/antham/helm-backup/pull/31

PR #66, opened 2017-12-12
amx: theme amx-save-file

jpkotta:
amx-save-file is a cache of amx-data and amx-history

https://github.com/DarwinAwardWinner/amx


tarsius:
Assuming the file contains an s-expression the conventions call for it to be named `amx-save.el`.

tarsius:
Thanks!

I have squashed the commits.

PR #69, opened 2017-12-18
Add note about custom-file to README.md

hinrik:
See https://www.gnu.org/software/emacs/manual/html_node/emacs/Saving-Customizations.html.


tarsius:
Are you okay with this fixup being committed in your name?

hinrik:
Yeah, no problem.

PR #71, opened 2017-12-19
remember: theme remember-data-file and remember-data-directory

hinrik:
See:

https://www.gnu.org/software/emacs/manual/html_node/remember/Text-File.html

https://www.gnu.org/software/emacs/manual/html_node/remember/Separate-Text-Files.html


tarsius:
I have done this differently in 1cbf83a4c54e4262936cdba8f77b6f5bd3c7afc2. When there are multiple files/directories, then we use turn the prefix into a directory and we drop suffixes such as `-file` and `-directory`. Because in this case we end up with two identical names, so we have to add some other suffix. If the file did contain an s-expression, then I would have appended `.el` to the respective file. Because it does not I have appended `.d` to the directory instead.

PR #72, opened 2017-12-30
pyim: theme `pyim-directory' and `pyim-dcache-directory'

kaiwk:
pyim is a Pinyin input method.

- `pyim-directory' is used to store related files.
- `pyim-dcache-directory' will store dictionary cache.

See also: https://github.com/tumashu/pyim

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -324,6 +324,7 @@ directories."
       `(make-directory ,(var "projectile/") t))
     (setq projectile-cache-file            (var "projectile/cache.el"))
     (setq projectile-known-projects-file   (var "projectile/known-projects.el"))
+    (setq pyim-dcache-directory            (var "pyim/dcache/"))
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

Even though upstream doesn't do it for the default values, you should append a slash to the values to indicate that we are dealing with directories. You might also want to ask upstream to do this for the default values.

It appears that `pyim-directory` isn't actually used for anything. You might want to ask upstream about that.

tarsius:
While what you did is probably more faithful to the conventions, I think we should just go with ` (var "meghanada/") - putting the binary in some subdirectory seems like overkill.

kaiwk:
Thanks for the advice!

I should thought that before, since the package is not generally used by common Emacs users.

tarsius:
Thanks!

I have rebased this.

PR #73, opened 2017-12-31
meghanada: theme `meghanada-server-install-dir'

kaiwk:
meghanada is a Java completion engine.

This is the directory to store the meghanada server which will be conneced when
enable meghanada mode. When call `meghanada-update-server', it will download it
from 'dl.bintray.com'.

See also: https://github.com/mopemope/meghanada-emacs/blob/98ad6a5361c725319a355522d2d1ba0e0fbb7cde/meghanada.el#L107

Happy new year!

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -307,6 +307,7 @@ directories."
     (setq magithub-dir                     (var "magithub/"))
     (setq magithub-cache-file              (var "magithub/cache.el"))
     (setq mc/list-file                     (var "mc-list.el"))
+    (setq meghanada-server-install-dir     (var "meghanada/"))
     (setq multi-compile-history-file       (var "multi-compile-history.el"))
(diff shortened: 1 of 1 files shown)


tarsius:
While what you did is probably more faithful to the conventions, I think we should just go with ` (var "meghanada/") - putting the binary in some subdirectory seems like overkill.

tarsius:
Thanks!

I have squashed your commits. For such trivial iterations at least it is usually preferable to just amend to the existing commit instead of adding additional commits. If we should change our mind and want to go back to an earlier iteration, then we could still recover that from the reflog. I actually prefer that even for more complex changes, except that in such cases I usually create fixup commits (using `c f` in Magit) and then combine those commits into the "final commit" shortly before merging.

kaiwk:
I don't even know that, thank you!

PR #74, opened 2018-01-09
add support for keyfreq: keyfreq-file and keyfreq-file-lock

wbolster:
there are two file settings, but a subdirectory is not needed, since
in practice there is only one file (the lock file is not persistent).


tarsius:
I have adjusted the commit message to conform to the conventions.

Placing two files directly in `etc/` also violates the conventions, but I agree with your reasoning for violating it.

PR #75, opened 2018-01-16
nov: theme nov-save-place-file

dakra:
- This file is used to save and restore the last reading places.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -315,6 +315,7 @@ directories."
     ;; actual file name MUST end with ".el".  Use "git blame" for
     ;; more information.
     (setq notmuch-init-file                (etc "notmuch-init"))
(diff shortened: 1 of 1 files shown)


tarsius:
In what format is data stored in this file?

dakra:
> In what format is data stored in this file?

s-expression

I updated the commit message.

tarsius:
Then you should add the `.el` extension.

Please provide a link to the package repository so that I can check whether it makes sense to additionally violate the conventions by naming the file `nov-places.el` instead of `nov-save-place.el`.

dakra:
I put a link and description in the commit message and updated the variable name.
I was mistakenly just taking the default value. Don't think it warrants an exception so `nov-save-place.el` it is.

Thanks for the patience.

tarsius:
I agree.

Thanks!

PR #77, opened 2018-03-31
window-purpose: theme variables

TheBB:
Theme `purpose-default-layout-file` and `purpose-layout-dirs`.

purpose-default-layout-file:

purpose-layout-dirs:
- Files stored here are also used for s-expressions, but the package does not allow changing the
  file name extensions.

Upstream: https://github.com/bmag/emacs-purpose


tarsius:
Thanks!

PR #59, opened 2017-09-14
Auto package update: Theme last update day file

YourFin:
Theme `apu--last-update-day-path'.
This file is used to store raw text and is the only data file of the package.

[Auto Package Update repo](https://github.com/rranelli/auto-package-update.el)

PR #80, opened 2018-04-12
auto-package-update: theme `apu--last-update-day-filename`

andschwa:
This file is used to store raw text.
This is the only configuration/data file of the package.
Its default value is `.last-package-update-day`.

https://github.com/rranelli/auto-package-update.el

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -250,6 +250,7 @@ This variable has to be set before `no-littering' is loaded.")
     (setq ac-comphist-file                 (var "ac-comphist.el"))
     (setq amx-save-file                    (var "amx-save.el"))
     (setq anaconda-mode-installation-directory (var "anaconda-mode/"))
+    (setq apu--last-update-day-filename    (var "auto-update-package-last-update-day")
     (setq async-byte-compile-log-file      (var "async-bytecomp.log"))
     (eval-after-load 'bbdb
       `(make-directory ,(var "bbdb/") t))


andschwa:
Oops, were we not supposed to put backticks around the variable name in the commit message?

tarsius:
That's not a big deal. What's more problematic is that you didn't test this at all - a closing parenthesis is missing.

And I am afraid I don't really want to add this package. There was another pr earlier (#59), which lead to a pr asking for the option `apu--last-update-day-filename` to be renamed (https://github.com/rranelli/auto-package-update.el/issues/31), but nothing has happened there yet.

Because you have themed `apu--last-update-day-path` instead I noticed that the problems runs deeper. Customizing `apu--last-update-day-filename` after the library has already been loaded won't have an effect because that won't cause the value of  `apu--last-update-day-path` to change. (Which I assume is the reason you themed the latter.)

These issues have to be fixed before I can merge a pr theming this package. The option should be renamed to `auto-package-update-last-update-day-file` (or something shorter), the variable should be removed, and every function that used to use it should do the `expand-file-name` itself.

andschwa:
Oops, my bad. Sorry about that @tarsius. I owe you more beer.

tarsius:
> I owe you more beer.

Or figure out the name of the mentioned function for me :stuck_out_tongue_winking_eye:

PR #81, opened 2018-04-14
Clarify the roles of etc/var directories

### LINKED ISSUE #78: There is no explanation behind var and etc logic
- What exactly should go into var?
- What exactly should go into etc?

Fuco1:
I think repeating the logic makes it a lot clearer.

I get it that we are following the unix/linux conventions but sometimes I have doubts what should go where.  Are bookmarks configuration or data? I think it's both.  Do I want to version bookmarks? Do I want to version `var/` or only `etc/` or neither?  These are some additional questions I have that this PR does not address.  Do we have good answers?

Fix #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,11 @@ Furthermore sometimes file names are used that don't provide any
 insight into what package might have created them.
 
 This package sets out to fix this by changing the values of path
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks! I did refill a bit before merging.

PR #82, opened 2018-06-20
pandoc-mode: theme `pandoc-data-dir'

Guilz:
pandoc-mode repository: https://github.com/joostkremers/pandoc-mode

The `pandoc-data-dir` directory really contains config, not data, even though the variable is named `pandoc-data-dir`. The config files contain S-expressions describing the desired options for pandoc output formats.


tarsius:
Thanks!

PR #83, opened 2018-06-20
auto-package-update: theme variables

Guilz:
Repository for `auto-package-update`: https://github.com/rranelli/auto-package-update.el

Theme:

- `apu--last-update-day-filename`
- `apu--last-update-day-path`

The variable ending in `-path` actually also points to the file. This file contains the timestamp of the last automatic package update.

Without this `no-littering` rule, `auto-package-update` will put this file in `~/.emacs.d/.last-package-update-day`.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -251,6 +251,8 @@ This variable has to be set before `no-littering' is loaded.")
     (setq amx-save-file                    (var "amx-save.el"))
     (setq anaconda-mode-installation-directory (var "anaconda-mode/"))
     (setq async-byte-compile-log-file      (var "async-bytecomp.log"))
+    (setq apu--last-update-day-filename    (var "auto-package-update-last-update-day"))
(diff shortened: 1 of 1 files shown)


tarsius:
See #59, #80 and https://github.com/rranelli/auto-package-update.el/issues/31.

Guilz:
Oops sorry, I should have done some research before submitting this. 😕
Next time I will check the previous PR.

Guilz:
Now that rranelli/auto-package-update.el#32 is merged, should we reopen this PR?
I know you wanted the `apu-*` variables renamed with the same prefix as the rest of the package (`auto-package-update`), but as it is now that would already work ok.

I updated my fork accordingly, see Guilz/no-littering@a3f3955b82b7c47fbd39a7045897a2982ff203f0 but because this PR is closed it didn't pick up the new commit.

PR #84, opened 2018-06-26
emms: Automatically create emms folder for emms-directory

dakra:
Emms does *not* create automatically the folder for `emms-directory`.

I see popular package like `helm` and `elfeed` have
also `make-directory` functions in `no-littering` which makes me
think it is ok.

But then this confuses me a bit:

Should I fix this upstream?
If so, you can close this PR and I'll submit a patch to emms instead.

Thanks.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -277,6 +277,8 @@ This variable has to be set before `no-littering' is loaded.")
 This function overrides the one on `x-win' to use `no-littering'
 directories."
              (expand-file-name session-id ,session-dir)))))
+    (eval-after-load 'emms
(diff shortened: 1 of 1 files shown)


tarsius:
> But then this confuses me a bit: [...]
> Should I fix this upstream?
> If so, you can close this PR and I'll submit a patch to emms instead.

Well, it seems fairly clear to me: *also* ask upstream to create the directory (in addition to adding a kludge here). I think Emms is actively maintained, so changes are good such a change will make it in a timely manner. I'll leave this open for a few days. If nothing happens upstream, then please ping me and I will merge this.

PR #87, opened 2018-08-25
hackernews: theme hackernews-visited-links-file

basil-conto:
This file is used to store an s-expression.  This is the only data file of the package to date, but the package places the file in a subdirectory of its own by default, to accommodate future files.  This package takes care of creating the containing directory if necessary.

Homepage: https://github.com/clarete/hackernews.el


tarsius:
Excellent commit message!

basil-conto:
Excellent commit instructions! :)

PR #92, opened 2019-02-26
logview: theme logview-cache-filename and logview-views-file

dakra:
logview-cache-filename:
- This file is used to store cache data

logview-views-file:
- This file is used to store different views (/filters)

Repo link: https://github.com/doublep/logview

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -314,6 +314,8 @@ directories."
     (eval-after-load 'lookup
       `(make-directory ,(etc "lookup/") t))
     (setq litable-list-file                (var "litable-list.el"))
+    (setq logview-cache-filename           (var "logview-cache"))
+    (setq logview-views-file               (etc "logview-views"))
     (setq lookup-init-directory            (etc "lookup/"))
     (setq lsp-session-file                 (var "lsp-session.el"))
     (setq magithub-dir                     (var "magithub/"))


no-littering.el
```diff
@@ -314,6 +314,8 @@ directories."
     (eval-after-load 'lookup
       `(make-directory ,(etc "lookup/") t))
     (setq litable-list-file                (var "litable-list.el"))
+    (setq logview-cache-filename           (var "logview-cache.extmap"))
```
tarsius: Why a suffix?

no-littering.el
```diff
@@ -314,6 +314,8 @@ directories."
...
+    (setq logview-views-file               (etc "logview.views"))
```
tarsius: Should be `"logview-views"`.

no-littering.el
dakra: I simply put it there because the original value had it. Changed now.

no-littering.el
dakra: Thanks. Both fixed.

tarsius:
Did you check whether the package takes care of creating the directories if they don't exist? (Same for your other recent changes.)

dakra:
> Did you check whether the package takes care of creating the directories if they don't exist? 

They're only a single file each and don't create any directories.

> Same for your other recent changes
`dab-breakpoints` is also a single file. `lsp-java` takes care of the directories. I'll mention it in the commit message for that.

PR #93, opened 2019-02-26
lsp-java: theme variables

dakra:
Theme `dap-java-test-runner', `lsp-java-workspace-dir'
and `lsp-java-server-install-dir'.

dap-java-test-runner:
- This file specifies the test-runner that is used in dap-mode

lsp-java-workspace-dir
- LSP java workspace directory

lsp-java-server-install-dir
- Install directory for eclipse.jdt.ls-server

Repository URL: https://github.com/emacs-lsp/lsp-java


tarsius:
Are the noisy paths actually needed?  The doc-strings mention that this is so.  If not, then it would be better to conform to the conventions of this project.  Who is responsible for putting these files there?

dakra:
I guess with noisy you mean you would remove the `eclipse.jdt.ls` part of the path from the test-runner and server-install-dir?!

The files are automatically put there if not already existing. But I'm just experimenting with `lsp-java` and didn't use it very much.

Maybe @yyoncho can give a better answer if that folder structure is really needed (I would guess it is not).

yyoncho:
`lsp-java` automatically downloads several files needed by lsp-java, lsp-java-boot, or dap-mode and each one of them is located under `~/.emacs.d/eclipse.jdt.ls/server/` . I am not sure what are the concerns with that folder structure but in general, any folder structure will work...

tarsius:
> I guess with noisy you mean you would remove the `eclipse.jdt.ls` part of the path from the test-runner and server-install-dir?!

Yes.

> The files are automatically put there if not already existing. But I'm just experimenting with `lsp-java` and didn't use it very much.

I am not even experimenting with it.  That means I do not want to do much investigating. I also don't want to install it for testing purposes.  You will have to do the work.  (And because this appears to be a somewhat complicated case and I only want to merge sufficiently documented changes, that might be a bit of work.  Alternatively feel free to not put in the necessary work and instead close this pull-request.)

But I had a quick look at the code at least.

`dap-java-test-runner` is not declared as a variable; its use is protected by `boundp`.  From the look of it I am guessing that this variable is obsolete and only respected.

(That's really something that you should have noticed and talked about in the commit message.  I.e. you need to justify why it is the appropriate thing to do to theme this variable anyway.)

But...

>`lsp-java` automatically downloads several files needed by lsp-java, lsp-java-boot, or dap-mode

... sounds like `lsp-java-server-install-dir` should be themed instead.

> and each one of them is located under `~/.emacs.d/eclipse.jdt.ls/server/`.  [...] any folder structure will work...

For `no-littering` the path should probably by something like `~/.emacs.d/var/lsp-java-server/`, but there are a few open questions here:  Probably `.../lsp-java/server` should be used, since this package has other data directories.  The `no-littering` conventions actually call for it to be named `.../lsp-java[-/]server-install`, but I *think* the "install" in the symbol name is probably unfortunate, it's a "install into *and then use from*" directory.  That's something that should be discussed with the upstream.  I might be completely wrong here, but it's the sort of thing that a contributor should think about.  Also all the path variables should be themed in one go.

> I am not sure what are the concerns with that folder structure but in general,

The main purpose of `no-littering` is to not litter `~/.emacs.d/` itself with lots of directories and files belonging to Emacs packages, i.e. to prevent `$HOME/.emacs.d/` from becoming as messy as `$HOME` itself.  This package "themes" the "path options" of other packages to prevent that from happening.

Since we do that anyway, we then also try to use consistent names for the new paths that we use.  For a package like `lsp-java` in their own right it may make sense to use a java-esk path.  But once `no-littering` themes it, consistency across all themed Emacs packages is more important than for each package using paths that are consistent with the non-Emacs thing it works with in some way.

dakra:
Thanks for the thoughtful comment.

> dap-java-test-runner` is not declared as a variable; its use is protected by
> `boundp`.  From the look of it I am guessing that this variable is obsolete and
> only respected for backward compatibility.

I missed that `dap-jva-test-runner` is *not* declared in `lsp-java`
but in `dap-mode` [dap-java](https://github.com/emacs-lsp/dap-mode/blob/master/dap-java.el#L48).

But `lsp-java` is actually doing the download and installation of
the test-runner.

> ... sounds like `lsp-java-server-install-dir` should be themed instead.

`lsp-java-server-install-dir` *is* themed?!

> For `no-littering` the path should probably by something like
> `~/.emacs.d/var/lsp-java-server/`, but there are a few open questions here:
> Probably `.../lsp-java/server` should be used, since this package has other data
> directories.  The `no-littering` conventions actually calls for it to be named
> `.../lsp-java[-/]server-install`,

For the java lsp server, the folder is
`lsp-java/eclipse.jdt.ls/server/`
and test runner:
`lsp-java/eclipse.jdt.ls/test-runner/junit-platform-console-standalone.jar`

You could argue it makes sense to have the extra eclipse folder
in case you want to install another java language server in
parallel?

> but I *think* the "install" in the symbol name
> is probably unfortunate, it's a "install into *and then use from*" directory.
> Since the "using" part is more important than the "installing" part (or at least
> more frequent), I think it makes little sense for "install" to appear in the
> name of the variable.  That's something that should be discussed with the
> upstream.  I might be completely wrong here (i.e. it may be a java thing to have
> "install" in the names of "location variables, I wouldn't know), but it's the
> sort of thing that a contributor should think about.

I can add another variable like `lsp-java-ls-dir` (naming?) upstream
that `dap-java-test-runner` and `lsp-java-server-install-dir` can use
so we have to only theme this one variable (and `workspace`).
Would this be better?


Thanks and sorry for the delay and not mentioning more info with the
PR.

tarsius:
I think this is good enough now and that improving it would require me to spend too much time learning about something I am not actually interested in. So I am merging as-is.

PR #95, opened 2019-06-17
Move `treemacs--last-error-persist-file` to corresponding directory

cireu:
This file is used to store org style file.

https://github.com/Alexander-Miller/treemacs/blob/master/src/elisp/treemacs-persistence.el#L37


tarsius:
Please adjust this so that it conforms to the [Conventions](https://github.com/emacscollective/no-littering#conventions).

cireu:
> Please adjust this so that it conforms to the Conventions.

I have to claim that `treemacs--last-error-persist-file` should be set after loaded `treemacs-persistence` or it won't work as expected.

Should make a new directory "treemacs" and move both `treemacs-persist-file` and `treemacs--last-error-perist-file` into it? It may break old behaviours.

tarsius:
> I have to claim that `treemacs--last-error-persist-file` should be set after loaded `treemacs-persistence` otherwise it won't work as expected.

That variable should be defined using `defvar` instead of `defconst` Maybe even `defcustom` should be used and the private prefix be dropped.

/cc @Alexander-Miller Other people might also like to change the location and/or name of this file.

Alexander-Miller:
On it.

Alexander-Miller:
Made the changes with https://github.com/Alexander-Miller/treemacs/commit/d836da4acd8b04b70cd9facf6c6e8fd4fc1ad706

cireu:
What about the compatiblity issue about moving both files to `treemacs` directory?

tarsius:
> What about the compatiblity issue when moving both files to `treemacs` directory?

Just move it. This package would be too complicated if it also had to deal with moving files. Maybe once 1.0 is releases.

cireu:
Done. @tarsius

tarsius:
> On it.

While you are at it, could you make sure that the code that saves to these files makes sure that the containing directory exits, basically `(make-directory (file-name-directory FILE) t)`. Thanks!

Alexander-Miller:
Done as of https://github.com/Alexander-Miller/treemacs/commit/a5329dd60f954d4178540607ea1fccefdaa97cef.

tarsius:
@cireu Please squash that into a single commit.

cireu:
Done.

tarsius:
I have changed the message a bit before merging.

Thanks to both of you!

PR #98, opened 2019-07-29
x86-lookup: theme x86-lookup-cache-directory

hogand:
This directory is used to store an s-expression file generated from
the PDF.  The generated file is named <sha1>_v3 in this directory.

The package will create the directory if it does not exist.

https://github.com/skeeto/x86-lookup


tarsius:
Thanks!

PR #99, opened 2019-08-08
Theme bookiez and debbugs

stepnem:
In both cases the file in question is the only file variable used by
the library.

debbugs-gnu has an undocumented half-implemented feature of saving the
bug list cache to a hard-coded file path (`debbugs-gnu-save-cache`),
but there is currently no way to load the cache back from the file and
the functionality doesn't seem very useful (it's stashed at the very
end of the source file), so instead of perhaps finishing it, making
the path customizable and submitting a patch upstream I decided to
ignore it.


tarsius:
So summer was nice and all but now you feel like sitting on your computer again? :stuck_out_tongue_closed_eyes:

stepnem:
On Thu, 08 Aug 2019 12:02:32 -0700
Jonas Bernoulli wrote:

> So summer was nice and all but now you feel like sitting on your
> computer again?

Yeah well, it's... complicated.

I've had the bookiez modification sitting around for 2 years, but only
with the recent addition of debbugs has it reached critical mass :-D

PR #100, opened 2019-08-11
prescient: Theme prescient-save-file

akirak:
This adds support for [prescient.el](https://github.com/raxod502/prescient.el) written by @raxod502.

This seems to be the only file variable in the package.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -347,6 +347,7 @@ directories."
     (setq persp-save-dir                   (var "persp-mode/"))
     (eval-after-load 'projectile
       `(make-directory ,(var "projectile/") t))
+    (setq prescient-save-file              (var "prescient-save.el"))
     (setq projectile-cache-file            (var "projectile/cache.el"))
     (setq projectile-known-projects-file   (var "projectile/known-projects.el"))
     (setq purpose-default-layout-file      (etc "window-purpose/default-layout.el"))


raxod502:
So the default value of `prescient-save-file` is `(expand-file-name "var/prescient-save.el" user-emacs-directory)` already, because I think no-littering conventions are a good idea. This of course doesn't handle `no-littering-var-directory`. Is there an official recommendation for how I as a package author should conform to the conventions, so that code in no-littering isn't necessary? Or is the goal to have every package themed in no-littering, just for consistency?

akirak:
The **default** value of `no-littering-var-directory` is `(expand-file-name "var" user-emacs-directory)`, but the package allows the user to customize the variable. Then the location will be different. I customized the variable for no reason, but maybe I should revert it back to the default value. 

Still, I don't want to put the etc directory in `user-emacs-directory` and will continue to alter the value of `no-littering-etc-directory`, so I wish to have variables of such packages themed in `no-littering`. This doesn't apply to prescient, since it doesn't have any `etc` files.

> Or is the goal to have every package themed in no-littering, just for consistency?

It will be a neat idea for individual package developers to follow the conventions of `no-littering` so that the maintainer of this package won't have to theme variables in every package. Is there a former discussion on sharing the conventions among package developers?

akirak:
I'll close this PR, since I don't have to get it merged any more.
Thank you for the advice.
If anyone is interested, please reopen the PR.

tarsius:
> Or is the goal to have every package themed in no-littering, just
> for consistency?

The long-time goal is that `no-littering` becomes unnecessary because Emacs itself provides the equivalent of `no-littering-{etc,var}-directory` and `no-littering-expand-{etc,var}-file-name`, and all built-in packages use these functions when appropriate.

I don't feel like pushing for such a change right now.  If you want to help with this you could start thinking about how to convince the Emacs developers.

You could also think about the issues and benefits that arise if something like this is implement as part of Emacs. Would replacing `locate-user-emacs-file` with two functions be a good approach...

> Is there an official recommendation for how I as a package author
> should conform to the conventions, so that code in no-littering
> isn't necessary?

I would recommend that you do *not* follow the `no-littering`-conventions by default.  Users who like those conventions already use `no-littering` (assuming they know about it) and for those users that use your package but not `no-littering` following the convention just means that your package is yet another package that puts its file in some weird place.

In other words I think it is better if individual packages continue to dump their files directly in `~/.emacs.d` because everyone else is doing that.

PR #101, opened 2019-10-09
org-journal: Theme org-journal-cache-file

akirak:
Please use a dedicated feature branch.  This is why:

     https://github.com/magit/magit/wiki/Dedicated-pull-request-branches

Please try to follow the conventions.

     https://github.com/tarsius/no-littering#conventions

In the past the majority of contributors have ignored at least some of the conventions.  Others did not fully understand them or had a good reason to depart from the conventions but did not explain why that is so. Unfortunately it is hard for me as the maintainer to tell whether a contributor did not invest enough time to get things right or just forgot to be explicit about their thought process. The result is that a pull request is actually more work for me than a simple "please theme PACKAGE from URL".

Going forward contributors are expected to follow the conventions more closely from the get-go and to be explicit about their thought process. Adding such statements to commit messages, would be helpful for example:

Also please link to the repository of the package that your pull request is theming.

Thanks!

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -341,6 +341,7 @@ directories."
       `(make-directory ,(var "org/caldav/save") t))
     (setq org-caldav-backup-file           (var "org/caldav/backup.org"))
     (setq org-caldav-save-directory        (var "org/caldav/save"))
+    (setq org-journal-cache-file           (var "org/journal-cache"))
     (setq pandoc-data-dir                  (etc "pandoc-mode/"))
     (setq pcache-directory                 (var "pcache/"))
     (setq persistent-scratch-save-file     (var "persistent-scratch.el"))


tarsius:
To me it looks like there is a bug in `org-journal-serialize`.  If the file isn't writable, e.g. because the parent directory does not exist, then it silently fails to serialize.  In our case that would happen when if no other `org` related package already saved a file in the same directory.  Please have a look and if appropriate create a pr to fix that.  This function should create the directory if necessary.

akirak:
Thanks for your review. I'm trying to patch the package.

akirak:
It has been merged, so this PR is ready for merging. Thanks again.

tarsius:
Thanks!

tarsius:
Quickly force-pushed to add the `.el` suffix.

PR #102, opened 2019-10-09
rime: Theme liberime-user-data-dir

akirak:

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -319,6 +319,7 @@ directories."
     (setq keyfreq-file-lock                (var "keyfreq.lock"))
     (eval-after-load 'lookup
       `(make-directory ,(etc "lookup/") t))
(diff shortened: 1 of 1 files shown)


tarsius:
Are we talking about https://gitlab.com/liberime/liberime/blob/master/liberime-config.el? If so then that should be mentioned in the commit message because this package isn't available from Melpa. Please add this to Melpa--I would rather not add settings for packages that aren't even on Melpa.

> The package name is 'liberime', but it is a thin wrapper around rime dynamic module, so I think 'rime' is a suitable name.

For consistency, lets go with `'liberime`.

akirak:
Thanks for your review. Yes, I am talking about the package, but it seems difficult to add it to MELPA --- The author has been inactive for several months, and the package doesn't have linting. For now, I will give up on this patch. I will consider reopening this issue when the author becomes active again.

PR #103, opened 2019-10-09
org-recent-headings: Theme org-recent-headings-save-file

akirak:


tarsius:
Why did you not provide any descriptions for the three pull-requests that you just opened? Instead you just left the instructions that were intended for you in there. That's not very useful for me, because I have to read it to see if you did actually answer any of my questions inline before I can discard it.

tarsius:
Lets drop the `-save` suffix. Also at first sight this appears to contain lisp data and should therefore use the `.el` suffix (see the conventions).

akirak:
Sorry about that. I was in rush. Thank you for your feedback.

PR #107, opened 2019-11-07
lsp-julia: theme lsp-julia-package-dir

TheBB:
Package: https://github.com/non-Jedi/lsp-julia

- This is the only data directory of the package.
- It stores a downloaded LSP server.
- This package does take care of of creating the containing directory if necessary.

Also fixed a sorting issue.


TheBB:
Actually, forget this for now. Looks like it's not as trivial.

PR #112, opened 2020-01-13
lsp-mode: Theme lsp-server-install-dir

akirak:
Hello,

Since [lsp-mode](https://github.com/emacs-lsp/lsp-mode) now support automated installation of server executables, I have added this variable.

A question: `lsp-session-file` which is configured just below the added variable is for the same package. Should I change it to `lsp-mode/session.el` or keep it as it is?


tarsius:
I went with "lsp/server" and "lsp/session.el".

PR #113, opened 2020-01-23
nix-buffer: Theme nix-buffer-directory-name

akirak:
https://github.com/shlevy/nix-buffer

This is the only custom file/directory variable in the package. It is a directory that contains build results of Nix and possibly a file named "trusted-exprs".


tarsius:
Thanks!

PR #115, opened 2020-01-27
undo-fu-session: theme undo-fu-session-directory

leotaku:
Package: [ideasman42/emacs-undo-fu-session](https://gitlab.com/ideasman42/emacs-undo-fu-session)

* This is the only configuration/data directory of the package.
* The files are used to store s-expressions.
* This package does take care of creating the directory if necessary.

Please ignore the previous PR, I messed up. I guess that's what I get for trying to use the GitHub web UI for a quick edit instead of using Magit...


tarsius:
Thanks! And sorry for the delay.

PR #116, opened 2020-03-02
Fix omission from 79a7650 (create "lsp" directory)

stepnem:
The server subdirectory might have the same problem, but my usage does
not involve that one so I only encountered this one error and was too
lazy to test the other one as it seems to require 3rd party
executables.


tarsius:
Thanks!

PR #118, opened 2020-03-08
filesets: theme filesets-menu-cache-file

ksqsf:
- This file is used to store s-expressions
- This is a built-in feature


tarsius:
I put the `menu-` into the filename.

PR #122, opened 2020-03-08
undohist: theme undohist-directory

ksqsf:
- This is a directory, and the package takes care of creating it.
- undohist is a third party package available from MELPA ([git repo](https://github.com/emacsorphanage/undohist))


tarsius:
Thanks!

PR #124, opened 2020-03-27
auto-package-update: theme auto-package-last-update-day-path

liuyinz:


tarsius:
The file should be named `auto-package-update-last-update-day` to comply with our own conventions but also because "daily" does not mean "the day in which the last update was run" (taken from the doc-string of `auto-package-update-last-update-day-path`).

liuyinz:
sorry, I use my own file's name, fix it already.

tarsius:
Thanks!

Merged after squashing the commits.

PR #126, opened 2020-06-30
helm: theme helm-net-curl-log-file

manuel-uberti:
This var was recently added to log curl errors when
helm-net-prefer-curl is set to t.

See: https://github.com/emacs-helm/helm/commit/dee9e48fa76ad4a0d7b4ad1f6f3e7dd53e74d7cd


manuel-uberti:
Note that I kept the original file name and extension, but I can change them if you prefer something different.

PR #127, opened 2020-07-12
rcirc: theme rcirc-log-directory

condy0919:
- This is a directory, and the package takes care of creating it.
- Built-in package (rcirc).


tarsius:
Thanks!

PR #129, opened 2020-07-20
calc: theme calc-settings-file

okamsn:
This is a built-in package. 

- The file is used to store s-expressions.
- This is the only configuration/data file of the package that I know of. It is the only one listed in the Customization buffer.
- The file is by default stored at the top level of the user Emacs directory.

Thank you.


tarsius:
This should be "calc-settings.el".

okamsn:
The name has been corrected. Should anything else change?

tarsius:
Please squash the second commit into the first:

```
git reset --soft HEAD~
git commit --amend --no-edit
git push --force-with-lease okamsn HEAD
```

okamsn:
I followed your instructions and squashed the second commit.

tarsius:
Thanks!

PR #130, opened 2020-07-25
quickurl: theme quickurl-url-file

condy0919:
- Built-in package (quickurl).


tarsius:
It might be a little weird in this case, but the convention calls for "quickurl-url.el".

condy0919:
The default value of `quickurl-url-file` is `"~/.emacs.d/quickurls"`, so I
simplily add `.el` suffix

PR #131, opened 2020-08-05
equake: theme equake-persistent-display-file

emacsomancer:
Equake[1]  needs to be able to figure out what display to open on. When the
display value is not set in the environment, this task is complicated.  So
as one possible fallback, every time a graphical frame is opened we store
its display value in `equake-persistent-display-file` as a potential cue to
a likely display when the current display value is
unset/inaccessible (i.e. fallback to last known good value).

   - This is currently the only configuration/data file of the package.

[1] https://gitlab.com/emacsomancer/equake


tarsius:
Thanks!

PR #132, opened 2020-08-17
persist: theme persist--directory-location

cvdub:
- [persist.el](https://elpa.gnu.org/packages/persist.html) is a third party package providing variables which persist across sessions.
- persist--directory-location is the location of persist directory where the variables are stored.


tarsius:
After merging some additions I usually let it sit for a few days but when I finally tag a new release then I can be almost be certain a new pull-request is opened within a few hours. :grin:

tarsius:
Thanks!

PR #133, opened 2020-08-30
timeclock: theme timeclock-file

condy0919:
- Built-in package (timeclock).


tarsius:
The file should be named `timeclock`.

condy0919:
> The file should be named `timeclock`.

renaming done.

PR #136, opened 2020-09-20
Theme sly-mrepl-history-file-name and erc-log-channels-directory

leungbk:
* sly: theme `sly-mrepl-history-file-name`

The variable is defined here:
https://github.com/joaotavora/sly/blob/8e22c1f62667434ec1acd1e0c6b7ec1c22dc0958/contrib/sly-mrepl.el#L602

This package is a Common Lisp IDE, and one component of it is this REPL.

* erc: theme `erc-log-channels-directory`

ERC is an IRC client. Chat logs can be stored in this variable,
defined here:
https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/erc/erc-log.el#n146


tarsius:
These two are unrelated so it would have been better to use separate pull-requests.

erc: 

The name should be `"erc-log-channels"` to match the variable name. Also, are you sure this package does not define any other variables that control where data is saved? If there are others, then you should use `"erc/log-channels"`, even if you don't theme the others at this time.

leungbk:
It seems there's at least one other `erc` data directory, so I went ahead and followed your suggestion, and also added another commit for that `defcustom` in this PR.

> These two are unrelated so it would have been better to use separate pull-requests.

Sorry, I'll do separate PRs in the future.

Issue #135, opened 2020-09-20
Defaults for `abbrev-file-name` and `eshell-aliases-file`


leungbk:
Presently, `abbrev-file-name` is set to `/var/abbrev.el`, and `eshell-aliases-file` is determined by `eshell-directory-name` (which lies in `/var` also).

I think both `abbrev-file-name` and `eshell-aliases-file` are things that someone would want to check into version control, and should be somewhere in `/etc` for that reason.

``` diff
---
 no-littering.el | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/no-littering.el b/no-littering.el
index b60fbea..23c44b2 100644
--- a/no-littering.el
+++ b/no-littering.el
@@ -204,7 +204,7 @@ This variable has to be set before `no-littering' is loaded.")
 
 ;;; Built-in packages
 
-    (setq abbrev-file-name                 (var "abbrev.el"))
+    (setq abbrev-file-name                 (etc "abbrev.el"))
     (setq auto-insert-directory            (etc "auto-insert/"))
     (setq auto-save-list-file-prefix       (var "auto-save/sessions/"))
     (setq backup-directory-alist           (list (cons "." (var "backup/"))))
@@ -217,6 +217,7 @@ This variable has to be set before `no-littering' is loaded.")
     (setq diary-file                       (var "diary"))
     (setq ede-project-placeholder-cache-file (var "ede-projects.el"))
     (setq epkg-repository                  (var "epkgs/"))
+    (setq eshell-aliases-file              (etc "eshell/alias"))
     (setq eshell-directory-name            (var "eshell/"))
     (setq eudc-options-file                (etc "eudc-options.el"))
     (eval-after-load 'eww
-- 
2.28.0
```

tarsius:
I agree but before you open a pull-request, please use `"eshell/aliases"` to match the variable name. Also check and mention in the commit message what kind of data these files contain. If a file contains s-expressions then its name should end with `*.el`, otherwise it should not.

PR #138, opened 2020-09-23
eshell: theme eshell-aliases file separately from other eshell files

leungbk:
The aliases file stores user-defined aliases used in Eshell. Users
would typically be inclined to check them into version control as they
would with snippets.

The em-alias.el file normally resides inside of the directory
corresponding to eshell-directory-name, and does not assist in
generating an additional containing directory, so we must make that
directory ourselves.

https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/eshell/em-alias.el#n103

Related to #135.


tarsius:
Thanks for all!

PR #141, opened 2020-10-20
gdb-mi: theme gdb-window-configuration-directory

leungbk:
Emacs 28 introduced `gdb-window-configuration-directory`, which
corresponds to a directory storing window-configuration files for GDB.

See https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/progmodes/gdb-mi.el#n631.

Presently this appears to be the only directory `gdb-mi` needs. But it's not hard to imagine something like a directory containing files specifying (say) breakpoint locations, so I specified a child subdirectory for window configurations.


no-littering.el
```diff
@@ -232,6 +232,9 @@ This variable has to be set before `no-littering' is loaded.")
     (setq eww-bookmarks-directory          (var "eww/"))
     (setq filesets-menu-cache-file         (var "filesets-menu-cache.el"))
     (setq gamegrid-user-score-file-directory (var "gamegrid-user-score/"))
+    (eval-after-load 'gdb-mi
+      `(make-directory ,(var "gdb-mi/window-configurations") t))
```
tarsius, line 236: I've that is really currently necessary, then please ask them to fix that.

PR #144, opened 2020-11-26
bmkp: theme bmkp-bmenu-commands-file

coffeepenbit:
- This file is used to store s-expressions that recreate bookmark lists
- This package already has two themed variables

Repo: [https://www.emacswiki.org/emacs/bookmark%2b-bmu.el](https://www.emacswiki.org/emacs/bookmark%2b-bmu.el)

Please let me know if there's anything I didn't do appropriately.


tarsius:
Thanks!

PR #145, opened 2020-11-27
elfeed-score: theme `elfeed-score-score-file'

yantar92:
This file is used to store an s-expression.  This is the only
configuration/data file of the package.

Original package: https://github.com/sp1ff/elfeed-score/


tarsius:
I think we should use `(etc "elfeed/score/score.el"))`.

- `elfeed/....` because `elfeed-score` extends `elfeed`. See B.4.
- `elfeed/score/` because the other files in  `elfeed/` are also directories and mainly because `score-score.el` would be just to weird and just `score.el` would mean stripping everything except for the prefix.

It's a weird case, but I think that's the way.

PR #146, opened 2020-11-29
hyperbole: theme variables

condy0919:
Theme `hbmap:dir-user`, `hbmap:dir-filename`, `gbut:file` and `hibtypes-git-repos-cache`.

- `hbmap:dir-user` is a directory, the package can create it if necessary.
- `hbmap:dir-filename` is used to store an s-expression.
- `gbut:file` is used to store raw text.
- `hibtypes-git-repos-cache` is used to store an s-expression.
- third party package [hyperbole](https://github.com/rswgnu/hyperbole/)

The naming of hyperbole is too odd...

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -338,6 +338,12 @@ directories."
     (setq helm-github-stars-cache-file     (var "helm/github-stars-cache.el"))
     (setq helm-net-curl-log-file           (var "helm/helm-curl.log"))
     (setq historian-save-file              (var "historian-save.el"))
+    ;; hyperbole starts
+    (setq hbmap:dir-user                   (var "hyperbole/"))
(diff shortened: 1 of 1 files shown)


tarsius:
I would rather not theme this package. It's too... esoteric.

I have taken a look and think that by changing the values of these variables you break functionality by doing so (at least sharing between different users). I do not intend to try out this package to confirm whether that is really the case because (while it sounds interesting) it also comes with an excessive amount of compile warnings, which makes my highly uncomfortable. And on top of that it ignores the prefix conventions... there's probably more.

condy0919:
Agreed.

BTW, the context awareness `M-RET` in hyperbole is worth an installation.

PR #147, opened 2020-11-30
pdf-view-restore: theme `pdf-view-restore-filename'

yantar92:
This file is used to store an s-expression.  This is the only
data file of the package.

Package homepage: https://github.com/007kevin/pdf-view-restore/

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -382,6 +382,7 @@ directories."
     (setq org-registry-file                (var "org/registry.el"))
     (setq pandoc-data-dir                  (etc "pandoc-mode/"))
     (setq pcache-directory                 (var "pcache/"))
+    (setq pdf-view-restore-filename        (var "pdf-view-restore.el"))
     (setq persist--directory-location      (var "persist/"))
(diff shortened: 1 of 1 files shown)


tarsius:
The file in your previous pull-request was a special case. In this--quite normal--case, you should go back to following the conventions more closely, so there should be no duplicated part in the filename, just: `(var "pdf-view/restore.el")`.

And don't use any quoting in the commit summary line.  Only use that in the rest of the commit message, if appropriate.

Cheers!

yantar92:
> The file in your previous pull-request was a special case.

Got it.

> And don't use any quoting in the commit summary line. 

I was confused by the example in README.

```
foo: theme variables

Theme `foo-config-file', `foo-cache-directory',
and `foo-persistent-file'.
```

Should the quoting be not used at all in the commit or just in the summary line?

tarsius:
Just not in the summary line. Do quote in the rest of the commit message.

tarsius:
The setting was in the wrong group, see rule C.1.

I fixed that in this case and I looked at all the existing setting and noticed that I have made the same mistake myself too. :roll_eyes:

PR #149, opened 2020-12-22
package: theme package-quickstart-file

condy0919:
- Built-in package (package)


tarsius:
Thanks!

condy0919:
> I think a closing paren is missing here.

Yes, you're right..

PR #151, opened 2021-01-07
command-log-mode: theme clm/logging-dir

liuyinz:
[link] : [command-log-mode](https://github.com/lewang/command-log-mode)
   - Third-Party package
   - This is the only data directory of the package to save log files.
   - This package doesn't take care of creating the containing
     directory if necessary. 
   - variable name is abbreviation of package's name

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -288,6 +288,9 @@ This variable has to be set before `no-littering' is loaded.")
     (setq bmkp-last-bookmark-file          (var "bmkp/last-bookmark.el"))
     (setq bookiez-file                     (var "bookiez"))
     (setq cider-repl-history-file          (var "cider-repl-history.el"))
+    (eval-after-load 'command-log-mode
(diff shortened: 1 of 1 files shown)


tarsius:
> This package doesn't take care of creating the containing directory if necessary.

I was going to ask you to first try to get upstream to fix that, but looking at https://github.com/lewang/command-log-mode/issues/4 it seems it would at least take a very long time to get this fixed.

> This is the only data directory of the package

So we should use `(var "command-log-mode-logging/")`.

liuyinz:
I try to email the author, but no reply yet. 
I fix the path name and in the first commit, it seems I insert the code between the 'company package **eval-after-load** and **setq** S-expression by mistake. Fix it already.
Thanks to your patience !

tarsius:
Thanks.

I squashed two commits and changed the commit messages a bit.

PR #152, opened 2021-01-08
clarify that container defaults to `user-default-directory`

bard:
When first reading the README, I thought extra configuration would be needed on Linux because of conflicting defaults — the default user directory here is `~/.config/emacs`, so `no-littering`'s defaults of `~/.emacs.d/var/` and `~/.emacs.d/etc/` would need adjustment. After looking at the code, however, I realized that `no-littering` already places `var` and `etc` under `user-default-directory`, whatever that is. This change aligns documentation with what code is already doing (at the cost of slightly heavier prose).

No commit prefix as not related to a package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,10 @@ insight into what package might have created them.
 
 This package sets out to fix this by changing the values of path
 variables to put configuration files in `no-littering-etc-directory`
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

PR #153, opened 2021-01-17
org-roam: theme org-roam-db-location

DamienCassou:
URL of the repository: https://github.com/org-roam/org-roam

- This file is used to store an SQLITE3 database.
- This is the only configuration/data file of the package (as far as I know).
- This package takes care of creating the containing directory if necessary (in `org-roam-db`).


tarsius:
Thanks!

PR #154, opened 2021-01-17
Fix rendering of ~ in README.md

cpitclaudel:
Unquoted, `~` renders as strikeout, so the readme looks like this:

> (defaulting to "etc/" under `user-emacs-directory`, thus usually
> "~/.emacs.d/etc/") and persistent data files in
> `no-littering-var-directory` (defaulting to "var/" under
> `user-emacs-directory`, thus usually "~/.emacs.d/var/"), and

(I've read the conventions file, but I wasn't sure what template to follow given that this is not about a specific package)


tarsius:
Thanks!

PR #155, opened 2021-01-24
lsp-eslint: theme lsp-eslint-library-choices-file

DamienCassou:
URL of the repository: https://github.com/emacs-lsp/lsp-mode

- There is another configuration option in this package to store a file: `lsp-eslint-unzipped-path`. Nevertheless, I didn't create an `"lsp/eslint/"` directory to store both because this variable has a default value of `(f-join lsp-server-install-dir "eslint/unzipped")`and `lsp-server-install-dir` is already configured by no-littering.
- This package takes care of creating the containing directory thanks to my PR: https://github.com/emacs-lsp/lsp-mode/pull/2559.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -358,6 +358,7 @@ directories."
     (eval-after-load 'lookup
       `(make-directory ,(etc "lookup/") t))
     (setq lookup-init-directory            (etc "lookup/"))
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

PR #156, opened 2021-06-02
devdocs: Theme devdocs-data-dir

kljohann:
devdocs.el is a documentation viewer that uses `devdocs-data-dir` to store
documentation for offline use.  For each documentation, a nested directory is
created (e.g., `cpp` or `python~3.9`).  In addition, an index file (`docs.json`)
is kept there.  The directory is created if it does not exist yet.

URL: https://github.com/astoff/devdocs.el


tarsius:
Thanks!

PR #158, opened 2021-06-19
Theme url-history-file and url-cookie-file

leungbk:
Both files contain s-expressions, so we use extension `.el`.


tarsius:
So the goal of this is that the file-names should end if ".el"? If so, then the commit messages should say so.

Also I think it would be slightly nicer to write `(var "url/configuration/cookies.el")`.

leungbk:
> So the (only) goal of this is that the file-names should end if ".el", right? The _commit messages_ should mention that too.

Done.

> Also I think it would be slightly nicer to write `(var "url/configuration/cookies.el")`.

Done.

Issue #142, opened 2020-11-07
[Question] Is gnus-home-directory themable?


condy0919:
The default value of `gnus-home-directory` is `~/` which is not located at `~/.emacs.d/`.

tarsius:
All the variables whose default values depend on the value of this variable should be changed in the same commit; in our case without referring to the value of this variable. Some of these files should go into var, others into etc. This variable itself has to be set for the benefit of future variables. It should probably point to a directory inside etc.

tarsius:
Hm wait... maybe we should only theme `gnus-directory`. I partially confused the two.

In any case, all the relevant variables should be themed at the same time and the commit message should carefully describe the rational. Please do that and I will then decide if I agree, require more information, or have another proposal.

condy0919:
There are amount of variables of Gnus to be themed. Let me figure out all the variables..

fernandodemorais-jf:
Hello,

Sorry for commenting in a already closed issue, but I have some questions related to this topic.

Recently I've been thinking about how to theme Gnus relevant dirs properly.
By reading `no-littering.el` guidelines I came up with this:

```emacs-lisp
(setq gnus-home-directory                     (etc "gnus/"))
(setq gnus-directory                          (etc "gnus/"))
(setq gnus-init-file                          (etc "gnus/init.el"))
(setq gnus-startup-file                       (etc "gnus/startup/newsrc"))
(setq gnus-cache-directory                    (var "gnus/cache/"))
(setq message-directory                       (etc "message/"))
(setq message-signature-directory             (etc "message/signature/"))
(setq message-auto-save-directory             (var "message/auto-save/"))
```

Normally `gnus-directory` and `message-directory`, with their default values, creates the `~/News` and `~/Mail` folders, respectively.

So, my doubt is:

The correct paths, following `no-littering.el` specifications, is like above or is more like bellow?

```emacs-lisp
(setq gnus-home-directory                     (etc "gnus/"))
(setq gnus-directory                          (etc "gnus/news/"))           ; As the default.
(setq gnus-init-file                          (etc "gnus/init.el"))
(setq gnus-startup-file                       (etc "gnus/startup/newsrc"))
(setq gnus-cache-directory                    (var "gnus/news/cache/"))     ; As the default.
(setq message-directory                       (etc "message/mail/"))        ; As the default.
(setq message-signature-directory             (etc "message/signature/"))
(setq message-auto-save-directory             (var "message/mail/drafts/")) ; As the default.
[...]

PR #160, opened 2021-08-24
gnus: theme variables

condy0919:
Theme `gnus-dribble-directory', `gnus-init-file' and `gnus-startup-file'.

First attempt to theme Gnus. Related discussion #142


condy0919:
As stated in #142, `gnus-home-directory`, `gnus-directory` and
`gnus-cache-directory` contain the real content, so don't theme theme.

tarsius:
From https://www.gnu.org/software/emacs/manual/html_node/gnus/Startup-Files.html:

> Most common Unix news readers use a shared startup file called .newsrc.

So IMO we shouldn't theme that.

> You can turn off writing the .newsrc file by setting gnus-save-newsrc-file to nil, which means you can delete the file and save some space, as well as exiting from Gnus faster. However, this will make it impossible to use other newsreaders than Gnus. But hey, who would want to, right? 

So I would say that `no-littering` users who use `gnus` exclusively shouldn't change the location of that file but disable using that file.

condy0919:
Updated. Don't theme the newsrc file.

Thaodan:
> 
> So IMO we shouldn't theme that.
> 
> > You can turn off writing the .newsrc file by setting gnus-save-newsrc-file to nil, which means you can delete the file and save some space, as well as exiting from Gnus faster. However, this will make it impossible to use other newsreaders than Gnus. But hey, who would want to, right?
> 
> So I would say that `no-littering` users who use `gnus` exclusively shouldn't change the location of that file but disable using that file.

What about `newsrc.eld`?

PR #162, opened 2021-11-03
conventional-changelog: Theme conventional-changelog-tmp-dir

liuyinz:
link: [emacs-conventional-changelog](https://github.com/liuyinz/emacs-conventional-changelog)

   - This is the only data file of the package.
   - This package does take care of creating the containing directory if necessary.


tarsius:
Thanks!

PR #165, opened 2021-11-08
org: theme org-persist-directory

meliache:
As of the latest org-mode main branch (post 9.5, from commit [3b48224](https://git.savannah.gnu.org/cgit/emacs/org-mode.git/commit/?id=3b48224994e821b5b6c4e3a603b3f1d28432f205)) org-mode introduces the variable `org-persist-directory` as part of the new `org-persist` package. It contains index files for the `org-persist` cache.

The original default directory base-name is `org-persist` and by default it is located in `(xdg-cache-home)` if that exists, or otherwise in `user-emacs-dir`.

I'm not sure about this PR because

1. This is not in an org release yet and the details such as the variable name and default locations might be subject to change

2. The default linux location in `xdg-cache-home` isn't really littering up my `user-emacs-dir` and maybe that was chosen for a good reason. However, I have seen other caches in `var` and as a `no-littering` user have gotten accustomed to expect them there

Convention-wise I wasn't sure whether to use `var/org/org-persist` or `var/org/persist` for the directory path, but I chose the latter because this seems what other org packages are using.


tarsius:
Thanks!

Works for me, but if anything should change, then please follow up with another pull-request.

PR #168, opened 2022-01-09
code-review: Theme code-review-download-dir and code-review-log-file

leungbk:
https://github.com/wandersoncferreira/code-review/

code-review allows for code reviews to be performed within Emacs.

code-review-download-dir is a directory into which binary files are
downloaded. This directory is made by the code-review package if it
does not already exist.

code-review-log-file is a log file to which the code-review package
outputs certain warnings. The warnings are not in the form of
s-expressions.


tarsius:
The conventions call for

```
    (setq code-review-download-dir         (var "code-review/downloads/"))
    (setq code-review-log-file             (var "code-review/log"))
```

Any reason to diverged from that? (Asking because I can see that you *did* read the instructions, but it appears you missed the parts about how to name the files.)

Such as, for example! the `s` at the end of `"code-review/downloads/"`, which diverges from the conventions. I decided to add it ayway because it's still close enough to our convention and "download directories" of all sorts usually use the plural.

leungbk:
> Any reason to diverged from that? (Asking because I can see that you _did_ read the instructions, but it appears you missed the parts about how to name the files.)

Sorry, I missed the relevant parts of the instructions. I updated the commit per your suggestions.

I also added another defcustom I missed: `code-review-db-database-file`. I chose to name it `database.sqlite` instead of `db-sqlite` or similar even though it is in `code-review-db.el`, since I think full names are easier to read than abbreviated forms.

PR #170, opened 2022-01-29
rime: theme rime-user-data-dir

laampui:
Theme rime-user-data-dir.

emacs-rime is a package for CJK(chinese, japanese, korean) users to input their characters, and `rime-user-data-dir` stores the config files.

https://github.com/DogLooksGood/emacs-rime
https://github.com/DogLooksGood/emacs-rime/blob/5c2ade217134f6f20cd405447af20825e5b44513/rime.el#L356


tarsius:
Thanks!

Issue #79, opened 2018-04-11
Migration guide/script


Fuco1:
Suppose I have my own setup with a bunch of custom paths configured with emacs customize interface.

Is there some way I could start using this package while keeping all the data files? I.e. I imagine some feature that would walk all the definitions, look for the current values and *move* the files over to the expected locations (set by this package).

This should probably be done while emacs is shut down to prevent weird caching issues.

It should also remove all the customizations because they would probably conflict with the settings in this package.

Does this make sense? Is it safe?

tarsius:
I don't see how we could easily and safely do that. Currently we *essentially* have an alist that maps variables to the intended values. But actually it is code, which we would have to parse somehow or convert to data permanently.

Alternatively a script could be provided and maintained separately, but a fear that would quickly end up being outdated.

When I wrote this I have knowingly decided against writing some migration script. I found that I didn't care much about preserving the persistent data -- just using Emacs regenerated it over time. And `~/.emacs.d` ... well that doesn't even exist in my case.

So yes, having a migration tool would be nice, but I think it would be to hard to get it right.

tarsius:
Closing because I don't think this is feasible.

PR #176, opened 2022-05-09
Semi-automatic migration guide

Fuco1:
Couple years ago I opened #79 and asked for migration.  Today I bit the bullet and wrote some code to support semi-automatic migration.  The high-level overview is:

1. Parse the `no-littering.el` source code for the themed variable.
2. Save the current sessions values
3. Load no-littering
4. Compare saved values with new values and generate a migration script
5. Users goes over the list and executes sexps one by one to move files to the new location
6. Delete the `customize-variable` customizations

The process is described in `migrate.org` which also contains the code which can be tangled and loaded (only defuns, no side-effects). 

I have now successfully migrated my setup using this guide.


tarsius:
Thanks!

I have indented the code-blocks and added a few more line breaks.

PR #189, opened 2022-08-27
harpoon: theme harpoon-cache-file

leungbk:
https://github.com/otavioschwanck/harpoon.el.git

This is actually a
directory (https://github.com/otavioschwanck/harpoon.el/blob/b08d4af6e4ab404c8a1031ff8cbfff49d5c8aec4/harpoon.el#L135-L136);
the actual file name (whose name is not configurable), relative to this directory, is "harpoon".

As a consequence of this naming, the cache file will end up having
path "[...]/var/harpoon".

This is the only data directory of the package.

The package creates this directory, as well as the "harpoon" file.


tarsius:
I've done it differently, see the commit.

PR #190, opened 2022-09-10
eshell: Theme `eshell-rc-script` separately from other Eshell files

okamsn:
This sets `eshell-rc-script` and `eshell-login-script`.  These script files are
the Eshell startup file, and do not contain data.  Instead, they contain a
series of Eshell commands, similar to `.bashrc`.

The moving of these files from `var` to `etc` follows the same reasoning used in
#138 for moving the Eshell aliases file.

The current configured values of these variables are `var/eshell/profile` and
`var/eshell/login`.  Following the conventions, they are changed to
`etc/eshell/rc` and `etc/eshell/login`, given that the suffix "-script" is
similar to a suffix "-file".


tarsius:
Thanks!

PR #194, opened 2022-10-27
gnus: Theme gnus-startup-file so newsrc.eld is themed

Thaodan:
- Theme gnus-startup-file' it was suggested that this variable
  shouldn't be themed however gnus own configuration `newsrc.eld` is
  based on `(concat gnus-startup-file ".eld")` whichm means themeing
  is the only way to theme `newsrc.eld`.

- There are no downsides on this except that gnus won't read other
  readers newsrc anymore which is suggested anyway.

Fixes: dcc96cbf5f018a91d406926d3b69715847ef665a
Re: #142


tarsius:
Makes sense.  Thanks!

PR #201, opened 2023-04-20
backup-directory-alist: Don't backup potentially sensitive data

### LINKED ISSUE #200: Unencrypted `password-store` entries are backed up
As noted [here](https://lonely.town/@wasamasa/110227701055899570), `pass` entries are backed up in plain text when `backup-directory-alist` is set to `nil`.

josephmturner:
Resolves #200

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -215,7 +215,11 @@ This variable has to be set before `no-littering' is loaded.")
     (setq abbrev-file-name                 (etc "abbrev.el"))
     (setq auto-insert-directory            (etc "auto-insert/"))
     (setq auto-save-list-file-prefix       (var "auto-save/sessions/"))
-    (setq backup-directory-alist           (list (cons "." (var "backup/"))))
+    (setq backup-directory-alist           (list
+                                            ;; Don't backup potentially sensitive information
+                                            (cons "^/dev/shm/" nil)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks! I didn't see that before pushing my own fix. I've chosen to not set the variable at all, since other $EDITOR users might open files elsewhere.

josephmturner:
Thank you!

josephmturner:
For future reference, you can keep more-or-less the previous behavior by ignoring `/dev/shm/` and `/tmp/` but backing everything else up in a single backup dir.
```
(setq backup-directory-alist (list
                              ;; Don't backup potentially sensitive information (pass stores plaintext in /dev/shm/ and /tmp/)
                              (cons "^/dev/shm/" nil)
                              (cons "^/tmp/"     nil)
                              (cons "."         (expand-file-name (locate-user-emacs-file "var/backup/")))))

```

tarsius:
I assume `pass` deletes the file it asked `$EDITOR` to edit, when done. But it does not know about the backup files, so those will persist until a reboot. So this is still not as safe as disabling backups altogether with `(setq make-backup-files nil)`.

You could also use custom `backup-enable-predicate` function that returns nil for `/dev/shm`. The default does so for `/tmp`, but not this directory.

PR #203, opened 2023-05-03
Document and optionally reduce risk of leaking secrets

tarsius:
Please see the wip problem description at https://github.com/emacscollective/no-littering/blob/safer-backups/README.org#backup-files.

### DIFF
--- a/README.org
+++ b/README.org
@@ -51,6 +51,24 @@ variables before loading the feature.
 #+end_src
 
 *** Suggested Settings
+**** Auto-save, backup and undo-tree files
+
+Creating backups increases the odds that secret information, which
+would otherwise never be stored in clear text, is written to disk.
+Likewise backups increase the odds that leaked secrets persist longer.
+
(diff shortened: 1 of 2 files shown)


tarsius:
Upstream issues:

- https://debbugs.gnu.org/cgi/bugreport.cgi?bug=62991
- https://github.com/emacsorphanage/undohist/issues/14
- https://codeberg.org/ideasman42/emacs-undo-fu-session/issues/6

ideasman42:
I don't think this fix is correct. Although it boils down to emacs not having a good way to check if a file is a subdirectory of another (as far as I can see).

- `temporary-file-directory` may not have a trailing slash, in this case checking the prefix could falsely detect a temp directory e.g. `/tmp_simple_example`.
- The check may fail on case insensitive file systems (leaking details anyway).

This PR addresses the issues noted above https://codeberg.org/ideasman42/emacs-undo-fu-session/pulls/7

PR #208, opened 2023-06-06
shell-maker: Theme shell-maker-history-path

meliache:
The path under which the history file directories for each `shell-maker` implementation are stored. Notable shell-maker implementations are `chatgpt-shell` and `dall-e-shell` by the same developer and contained in the same [git repository](https://github.com/xenodium/chatgpt-shell). For example, their histories are stored in

```
  <shell-maker-history-path>
  ├── chatgpt
  │   └── history
  └── dall-e-shell
      └── history
```

By default, it takes the value of `user-emacs-directory`. I didn't look into the code but in my tests the directory was created if absent. The history-files are simple text files containing the user prompt histories.

The `chatgpt-shell` package also creates the alias `chatgpt-shell-history-path` to this variable.


tarsius:
Thanks!

PR #209, opened 2023-06-10
Improve migrate.org and mention it in the README

edgar-vincent:
I was surprised to see that the README stated that no-littering could not help the user migrate their files while, actually, there was a handy migration guide in the repository. Moreover, some steps in the guide seemed unclear to me. This PR thus:

* Mentions the `migrate.org` file in the README and in the comments at the top of `no-litter.el`.
* Improves the way the results of the Elisp Babel block containing the migration script generator are printed in the org file.
* Improves the wording and the spelling in that same file.

### DIFF
--- a/README.org
+++ b/README.org
@@ -24,9 +24,10 @@ this package depends on user contributions to accomplish decent
 coverage.  Pull requests are highly welcome (but please follow the
 conventions described below and in the pull request template).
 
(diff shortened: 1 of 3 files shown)


tarsius:
Thanks!

I've reworded/replaced the main commit.

PR #210, opened 2023-06-11
ement: theme ement-sessions-file

edgar-vincent:
This file is used to save Ement sessions in the form of an s-expression. Its default location is `~/.cache/ement.el`.
It is the only data file of the package. 

`ement` does not create the parent directory of `ement-sessions-file`. I will report this upstream.

Thanks!


tarsius:
Ping me once that has been implemented.

tarsius:
> Should I close this PR?

No. Usually we would just teach no-littering to create the directory instead.

However here this is not relevant. We place `ement-sessions-file` directly in `no-littering-expand-var-file-name`, which always exists when using `no-littering`. There is no intermediate that either the package, the user or `no-littering` would have to create. Look at the packages for which we *do* create directories explicitly to learn when that is appropriate/necessary.

I have edited the commit message to remove the remark about the package not creating a directory.

edgar-vincent:
Thank you very much for your detailed comment.

PR #212, opened 2023-06-20
emacs-gc-stats: Theme emacs-gc-stats-file

edgar-vincent:
This file is used to store an s-expression.
This is the only configuration/data file of the package. 
This package doesn’t take care of creating the containing directory if necessary, but since it is stored in `no-littering-var-directory' itself, this is not needed.

Upstream URL: https://git.sr.ht/~yantar92/emacs-gc-stats


tarsius:
Thanks again!

Just in case you are micro dosing me on purpose; if you have another half to full dozen themed packages in the pipeline, you can create a single pull-request for all that. (I might then drop one or two of those, and ask you to add a new pr after addressing my feedback. But generally it seems I can merge your additions as-is.)

edgar-vincent:
Haha, I'm not doing it on purpose, just doing things one at a time when I have a moment. I do have plans for other themes though, so I'll make sure to do as you instruct.

Thanks!

PR #213, opened 2023-07-10
tabspaces: theme tabspaces-session-file

gokuldas:
[Tabspaces](https://github.com/mclear-tools/tabspaces) leverages the tab-bar feature to create buffer-isolated workspaces called tabspaces. The tabspaces-session-file is used for persisting tabspace configuration across emacs restarts.  The default configuration is to save the session file as 'tabsession.el' in the user-emacs-directory. This is an s-expressions file.  This is the only data file written by the package.

The variable is defined in [tabspaces.el](https://github.com/mclear-tools/tabspaces/blob/9c0915e1b19a1bbc65a365be3f7910a15dba3ac4/tabspaces.el#L436)

- The PR is using a dedicated feature branch
- I have read the conventions and have followed it to the best of my knowledge
- The config has been tested and verified to work on my system in a fresh Emacs installation

**Review request:** The package code is an easy read. This is the only data file produced by the package as far as I can find from the code and the tests. However, a second verification is appreciated.

Thanks!


tarsius:
Thanks! Looks good.

PR #214, opened 2023-07-12
Drop "default-file" suffix from bookmark-default-file path

grtcdr:
Hi Jonas,

Thanks for your work on `no-littering`.

This pull request drops the `default-file` suffix from the `bookmark-default-file` path per the fourth convention found in the commentary.

I really think the filename should be plural but that breaks other conventions.


tarsius:
In 01693f864815ea7308a30f9497f14a33b68e3a4f from 2016 we switched from `bookmarks.el` to `bookmark-default.el`. That's only the fifth commit and the first by someone other than me. We were just working out the conventions at that time, see #5, #6, #7 and 745568bc76cb86f8a3319945f0e5556eabc0c747.

The convention that you mentioned was written down all the way back then. Maybe we simply overlooked that we were already using `bookmark-default.el` instead of `bookmark.el`, but I think it is more likely that we consciously decided to keep the "default" in there.

> (A.4) If the name of a path variable ends with `-file`, `-default-file`, `-directory`, `-default-directory`, or something similar, then that suffix is usually dropped from the file name.

So in this case we choose to continue to drop the `-file` suffix, instead of the longer `-default-file` suffix. One reason for that likely was to avoid having to use `bookmark.el`, instead of `bookmarks.el`, which would have meant violating the more important rule (A.2).

There are arguments for and against all the mentioned potential file names, and given that we have already been using it for seven years, I think we should just stick to `bookmark-default.el`. That has the considerable advantage that it won't break everyone's bookmarks, and it is also the name I tend to favor.

tarsius:
I'll leave it as-is. I'm also marking it as `wontfix`, so that I'll have another look, iff I ever decide to improve suboptimal but established paths en masse.

PR #221, opened 2023-11-15
treesit: Theme treesit.el variables

RangHo:
Theme `treesit--install-language-grammar-out-dir-history`, and `treesit-extra-load-path`.

- This directory contains tree-sitter grammar shared object artifacts (basically, ELF files).
- The installation function included in this package takes care of creating the directory.

----------

The `treesit` component introduced since Emacs 29 requires that you build _grammar libraries_ for each language you want to parse. Unfortunately, the default hard-coded search path for said libraries is `user-emacs-directory/tree-sitter`, littering our `.emacs.d` once again.

This patch reconciles this by:

1. specifying an alternative load path that is within the `var` directory (`treesit-extra-load-path`),
2. providing a default install directory for the `treesit-install-language-grammar` interactive function (`treesit--install-language-grammar-out-dir-history`).

### Oddities

Unlike specified in the README file, the directory is named `tree-sitter`, instead of `treesit`, since it contains shared object artifacts that technically can be used by the [legacy tree-sitter integration](https://github.com/emacs-tree-sitter/elisp-tree-sitter) as well.

Also, the variable `treesit--install-language-grammar-out-dir-history` is not yet present on Emacs 29.1.90, and the interactive function `treesit-install-language-grammar` _always_ litters the Emacs directory. The relevant patch was already made in April in commit emacs-mirror/emacs@c146bd893dfbc8a4c92f9d1f33def8f29e7ece1f, though.


tarsius:
Thanks!

"Merged". (Side-note:)

Unfortunately this pull-request asks for your `master` branch to be merged into ours.  It is better to use a dedicated feature branch instead, as [explained here](https://github.com/magit/magit/wiki/Dedicated-pull-request-branches).  It is too late for this pull-request now, but the next time you open a pull-request (in any repository really), then please use a feature branch.

<sub>This is a [canned response](https://help.github.com/en/github/writing-on-github/working-with-saved-replies); it may not apply 100% in this case.</sub>

leungbk:
This change causes issues for Nix users; those of us who use Nix to install our tree-sitter grammars have our Emacs' `treesit-extra-load-path` updated by Nix (using `add-to-list`) to contain `/nix/store/hash123-emacs-packages-deps/lib`: https://github.com/nixos/nixpkgs/blob/6a19af3e00770531136d8f27e3a452a549224bf4/pkgs/build-support/emacs/wrapper.nix#L166-L168

When I start up my Emacs, since this commit tosses out the Nix-prescribed update to `treesit-extra-load-path`, Emacs does not know where to find my pre-installed grammars.

Consequently, I am now using this in my config:

```elisp
(let ((old-val treesit-extra-load-path))
  (use-package no-littering
    :ensure t
    :demand t
    :config
    (setq treesit-extra-load-path old-val)))
```

Ideally, upstream Emacs would use some `defcustom` somewhere here instead of hard-coding: https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/treesit.el?h=ff1f82cbe3fa9aee354581f2798faaae7163ea44#n3424

tarsius:
Let's wait until this has been improved upstream. (And by "wait", I mean "feel free to bring this up upstream". :grinning: )

PR #222, opened 2023-12-13
racket: theme racket-repl-command-file

leungbk:
The file is basically a .bashrc (but in Racket) for racket-mode's newly redesigned
REPL mode.

The package creates the containing directory (https://github.com/greghendershott/racket-mode/blob/e149e3c920d63db6e0e9cc42f45c67f94a78a2b7/racket-repl.el#L525-L528).


tarsius:
Thanks!

PR #223, opened 2023-12-14
dape: Theme dape-adapter-dir

leungbk:
dape is a DAP client for Emacs.

At the time of writing, `dape-adapter-dir`, which is meant to be a
dumping ground for user-downloaded debug adapters, is the only
configuration file it defines.

`dape-adapter-dir` is not automatically created by `dape.el`; users
are expected to create `dape-adapter-dir` themselves before they
manually put their debug adapters there (unlike with `dap-mode`,
`dape` does not offer to download adapters for the user).


tarsius:
I think we should use `(var "dape-adapters/")`, see conventions A.4 and B.1. (There's nothing about the appended `s` yet, but I'll add that later.)

tarsius:
Thanks!

PR #225, opened 2023-12-18
erc: Theme erc-startup-file-list

leungbk:
erc-startup-file-list is a list of .bashrc-like files that are
searched for by erc, which executes the first extant and readable one.

If the executed file ends with a ".el" extension, then it is loaded as
an Elisp script, while if it does not end with that extension, then it
is treated as a bunch of IRC commands.

Following the upstream convention, we put the Elisp variant prior to
the non-Elisp variant in the list.


tarsius:
Following *our* conventions, this should be

```elisp
(list (etc "erc/startup.el") (etc "erc/startup"))
```

We should probably also keep `(... ".ercrc.el" ".ercrc")`, which I assume names files in `default-directory`.

PR #226, opened 2024-01-08
blamer: Theme blamer-avatar-folder

leungbk:
`blamer-avatar-folder` stores avatars corresponding to GitHub or GitLab
contributors.

-------

Presently, the package defines only a single variable corresponding to a data/config directory, but the name does not imply that it precludes the later introduction of other such `defcustoms`, so I followed rule 2.3.5:

> If a package only defines a single variable that specifies a data (or config) directory, then the directory name should nevertheless be just the package name. E.g., the path used for sx-cache-directory from the sx package is sx/cache/, not sx-cache/.


tarsius:
Thanks!

PR #228, opened 2024-02-01
gptel: Theme gptel-crowdsourced-prompts-file

yantar92:
This file is used to store csv table.
This is the only data file of the package.
This package doesn't yet (see https://github.com/karthink/gptel/issues/202) take care of creating the containing directory if necessary.


tarsius:
In that case we'll wait for upstream. (I would recommend opening a pull-request. The current issue IMO isn't explicit enough, it should mention that the package *should* create the directory. The way it is worded now, the initial response may very well be "yes, you have to create the directory yourself".)

yantar92:
Upstream resolved the problem.

PR #231, opened 2024-02-15
ielm: theme ielm-history-file-name

leungbk:
ielm-history-file-name is a variable new to Emacs 30. At the time of
writing, this is the only data file defined by IELM.


leungbk:
cf. https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=60cff1ac9d216e5abcb350ea5e623ab0b377c131

PR #232, opened 2024-02-27
tempel: Theme tempel-path to etc/

11xx:
Placing the [s-expressions lisp data file](https://github.com/minad/tempel?tab=readme-ov-file#template-file-format) in "etc/tempel-templates.eld" since it's a custom config file.

Just as a note there's also [tempel-collection](https://github.com/Crandel/tempel-collection) which provides *.eld 'template' files loaded from a directory (defined by `tempel-collection--dir`) but apparently they're not 'user-faced'.


tarsius:
Thanks!

I changed it to use a sub-directory, see commit for details.

PR #237, opened 2024-05-18
emacs: Tutorial state(s) directory

thoelze1:


thoelze1:
I mistakenly used `tutorial-directory` which is where the tutorial itself is stored (in its various languages). I should have used `tutorial--saved-dir` which yields the directory where saved states are kept. Unfortunately `tutorial--saved-dir` is a function intended for internal use ([manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Function-Names.html)) and not intended to be customized and so should stay out of `no-littering.el`. You can still customize it but you must require `tutorial.el` before redefining `tutorial--saved-dir` or else it will get clobbered:
```
(require 'no-littering)
(require 'tutorial)
(defun tutorial--saved-dir () (no-littering-expand-var-file-name "tutorial/"))
```
This works for me in my init file but I don't know enough to say whether it is technically undefined behavior.

Issue #240, opened 2024-07-20
[Feature Request] eFar file manager


AndersonTorres:
The package [eFar](https://github.com/suntsov/efar) provides a [FAR](https://github.com/FarGroup/FarManager)-like file manager for Emacs.

Bumping on it to experiment a little bit, I noticed it creates a hidden file, namely `.efar-state`. Digging the code, I found this line:

https://github.com/suntsov/efar/blob/78618a6cd9fe7d46c3728db3589d1fe50f7c1c6b/efar.el#L147

Is this a good opportunity to add eFar to no-littering?

tarsius:
Sure. Likely like so: `(setq efar-state-file-name (var "efar-state"))`. Please open a pull-request.

PR #241, opened 2024-07-20
efar: theme efar-state-file-name 

AndersonTorres:
Closes https://github.com/emacscollective/no-littering/issues/240


tarsius:
Thanks!

I had another look an noticed that the file contains a printed representation of a hash-table.  It is readable by `read`, or in other words it is lisp data, so I added the `.eld` suffix.

PR #246, opened 2024-09-18
gitignore: Exclude *-pkg.el

gs-101:
Exclude "*-pkg.el" files. These are generated by the `:vc' keyword from use-package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /*.elc
 /*-autoloads.el
+/*-pkg.el
 /.config.mk
 
 /docs/*.html


tarsius:
Please instead add that to `$XDG_CONFIG_HOME/git/ignore`. Doing so will save you a lot of time because most package repositories don't have this in their `.gitignore`.

tarsius:
I might merge this, but please don't take that as encouragement to open similar pull-request in my other repositories.

Like `*-pkg.el`, `*.elc` should IMO be ignored in the global ignore file, but over the years many users opened pull-requests like this one, so eventually I decided to just get it over with and add that to the `.gitignore` of all my packages.

But even if I end up deciding to do the same here, I want to do it on my own schedule, i.e., only when a package has other unreleased changes. I now release all packages monthly (or at least every other month), and I don't want to create releases just for this.

gs-101:
> Please instead add that to `$XDG_CONFIG_HOME/git/ignore`.

Thank you, that seems to be a better idea (I didn't know about this, sorry about that).

> I might merge this, but please don't take that as encouragement to open similar pull-request in my other repositories.

I'd say to not merge it and leave it closed as a sort of message or warning for other users.
I'd recommend adding a section/line mentioning the global ignore configuration to your  "contributions" pages instead, since users who are looking forward to contributing would check that first.

tarsius:
> I'd say to not merge it and leave it closed as a sort of message or warning for other users.

I'll close for now, but if I start getting as many requests for this as I got for `*.elc` and `*-autoloads.el`, then I'll add this rule too.

> I'd recommend adding a section/line mentioning the global ignore configuration to your "contributions" pages instead, since users who are looking forward to contributing would check that first.

That too, I would have to add that to all my packages, some of which don't even have a "contributions page". Adding a note there would be more annoying than just adding the rule. (More annoying for everyone: me, users who wanted the rule added, and users who couldn't care less but had to read about it anyway.)

PR #248, opened 2024-10-29
org-node-fakeroam: Theme variable org-node-fakeroam-data-dir

meedstrom:
Hallo.  I'm the maintainer of https://github.com/meedstrom/org-node-fakeroam.

I was not sure whether to put it in org/ or its own directory org-node-fakeroam/, but org/ seems to fit the prior art.

Trivia: this directory is likely to only ever contain one file, but it seemed more futureproof on my end to manage a dir name than a file name.


tarsius:
Thanks!

PR #250, opened 2024-12-01
nnmail: Theme nnmail-message-id-cache-file

Thaodan:
Theme nnmail's `nnmail-message-id-cache-file'.
Since the package is related to Gnus store the file in the Gnus
folder.

### DIFF
--- a/no-littering.el
+++ b/no-littering.el
@@ -145,6 +145,7 @@ This variable has to be set before `no-littering' is loaded.")
     (setq gnus-init-file                   (etc "gnus/init.el"))
     ;; Gnus hardcodes newsrc.eld to be based on gnus-startup-file.
     (setq gnus-startup-file                (etc "gnus/newsrc"))
+    (setq nnmail-message-id-cache-file     (var "gnus/nnmail-cache"))
     (setq ido-save-directory-list-file     (var "ido-save-directory-list.el"))
     (setq ielm-history-file-name           (var "ielm-history.eld"))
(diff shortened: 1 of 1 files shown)


tarsius:
The line is misplaced, since we sort by variable name, not file name.

Placing this file inside `gnus/` violates a convention, but that seems to make sense in this case. However if we violate that part of the conventions, then it becomes more important to follow other parts; so we should name the file after the variable name, i.e., `nnmail-message-id-cache`.

Then again, I took a quick look at other file variables inside `lisp/gnus/` and got the impression that it might make sense to relax the rules for "gnus et al. But if we do that, we should theme them all, starting with the two directory variables the values of other variables derive from. But I would want to think about that some more, and it's also more work.

But it is December, a month that somehow seems much shorter than the other months and yet I always have way more important things to finish than usual. In other words, let's instead look into theming all of gnus in late January.

Thaodan:
Sure I can fix the issues you mentioned.

I didn't pay attention to the ordering. What other Gnus variables are
not themed?

tarsius:
> Sure I can fix the issues you mentioned.

The thing is that I am *considering* going with what you have done here eventually, so I don't want to merge a later iteration that more closely follows the conventions. I am not sure though, I might eventually decide to strictly adhere to the conventions. I just don't know yet and don't want to merge this either way, because whatever we pick, I might later (when I have the time to think about it and experiment) change my mind.

> What other Gnus variables are not themed?

`gnus-home-directory` and `gnus-directory` and at least all the variables that use the values of these two when their own default values are set.

PR #251, opened 2024-12-23
eglot-fsharp: Theme eglot-fsharp-server-install-dir

RangHo:
This PR themes `eglot-fsharp-server-install-dir` variable.

The F# Software Foundation maintains their own [`fsharp-mode`](https://github.com/fsharp/emacs-fsharp-mode) for emacs, and it is shipped with a built-in support for `eglot` that automatically installs F# language server.

Theming this variable will keep the langserver binary out of the way.


tarsius:
I think `(var "eglot-servers/FsAutoComplete/")` or `(var "eglot-servers/fsharp/")` would make more sense. The idea is to put all lsp servers inside `eglot-servers`.

I'm unsure whether `fsharp/` or `FsAutoComplete/` should be used. Using the upstream name might be better because the user may install multiple competing servers for the same language. On the other hand, if I see "fsharp" I know what that is, while for "FsAutoComplete" I have no idea.

RangHo:
That also makes sense. I haven't seen other `eglot-*` packages auto-installing langservers so far, hence the initial commit.

I like the idea of keeping `(var "eglot-servers/<language name here>")` and let `eglot-*` do whatever they want within that directory. Since users install langservers themselves (by default), if a user were to maintain their own servers for a language, they can do it pretty easily (which, then, is no longer in our jurisdiction anymore).

One question though: in that case, should we name the collective directory `eglot-server-programs` (from `eglot-server-programs-alist`) or is that too verbose?

tarsius:
> One question though: in that case, should we name the collective directory `eglot-server-programs` (from `eglot-server-programs-alist`) or is that too verbose?

I've never used eglot, so I have some questions.

I considered using that name for that reason, but looking at the definition of this variable, I do not get the impression that it actually maps to directories like the one we are dealing with here. It maps major-modes to executables, which it names by name, without an absolute path. More or less.

It's under the control of other variables, such as the one we are dealing with here, where these executables are to be found.

Also, do this per-server directories contain just "a program" or is there usually a lot of support fluff as well? I guess it's the latter, so we should us "server" (as in "all the things the server needs"), not "server program" (as in "there's nothing but an executable").

RangHo:
> I do not get the impression that it actually maps to directories like the one we are dealing with here. It maps major-modes to executables, which it names by name, without an absolute path.

Correct, and that is also why I mentioned that it is easy for users to manage langservers themselves---they can just point to the right program or add the program to `PATH`.

Another accepted value is a function that *resolves* to a program name. Consider this example from [my Julia configuration](https://github.com/RangHo/dotfiles/blob/main/emacs/.config/emacs/config.org#julia):

``` emacs-lisp
(defconst rangho/julia-language-server-project
  (no-littering-expand-var-file-name "julia-language-server/"))

(defconst rangho/julia-language-server-script
  (expand-file-name "usr/bin/eglot-julia" user-emacs-directory))

(defun rangho/julia-mode-contact (_interactive)
  "Produce the Eglot contact information for Julia language server."
  ;; Make sure that the Julia project directory is available
  (unless (file-exists-p rangho/julia-language-server-project)
    (make-directory rangho/julia-language-server-project t))
  ;; Produce the contact information
  (list (executable-find "julia")
        "--startup-file=no"
        (concat "--project=" rangho/julia-language-server-project)
        rangho/julia-language-server-script
        (file-name-directory (buffer-file-name))))

(use-package julia-mode
  :init
  (add-to-list 'eglot-server-programs
               '(julia-mode . rangho/julia-mode-contact)))
```

The supporting packages like `eglot-fsharp` also work by adding these contact functions to `eglot-server-programs` alist. (See https://github.com/fsharp/emacs-fsharp-mode/blob/677d78c4d6cb574086408082dedbcaef04a85359/eglot-fsharp.el#L193-L202.)

In a nutshell, users usually install their own language servers (e.g. in `~/.local/bin`); not every language needs their own `(var "eglot-servers/<lang>")`. In this case, fiddling with .NET on Linux is super finicky, so the folks at the F# Foundation provided us with a set-it-and-forget-it solution for Eglot (which litters `.emacs`).

> Also, do this per-server directories contain just "a program" or is there usually a lot of support fluff as well? I guess it's the latter, so we should use "server" (as in "all the things the server needs"), not "server program" (as in "there's nothing but an executable").

That depends on the langserver implementation we're dealing with. For example, [Crystalline](https://github.com/elbywan/crystalline) is just a regular binary that we can drop wherever we want, while [Svelte language server](https://github.com/sveltejs/language-tools/tree/master/packages/language-server) is an entire NPM package with quite hefty dependencies.

Since we essentially are dealing with these "Eglot support packages", we *could* name them something like `(var "eglot-helpers/<language>")` too. I personally think all of `eglot-servers/`, `eglot-server-programs/`, `eglot-helpers/` make sense, though.

I guess it is up to you ;)

tarsius:
Thanks for providing these details.

> I personally think all of `eglot-servers/`, `eglot-server-programs/`, `eglot-helpers/` make sense, though.

I prefer `eglot[-/]servers/`. Because of conventions B.4 and B.5, we should go with `eglot/servers/`, not `eglot-servers/`. B.6 does not apply.

RangHo:
(Sorry, was out to grab some snacks)

I renamed the directory to `(var "eglot/servers/fsharp/")`, so that it is obvious what the directory is for.

I assume `eglot-foo` package will have full control over its own data directory `(var "eglot/servers/foo/")` from now on?

Let me know if you prefer `eglot/server/FsAutocomplete`!

tarsius:
> I assume `eglot-foo` package will have full control over its own data directory `(var "eglot/servers/foo/")` from now on?

I'm not sure what you mean by that. As long as neither a package such as `no-littering` nor the user themself have modified the location, the package's default choice is used. As before. Unless `eglot` decides to enforce something, but I would be surprised if it started to do that.

RangHo:
Got it. Well, I guess it also depends on the nature of each package. Please disregard that question.

I tested the latest version and it's working fine!

PR #256, opened 2025-06-09
server: Theme server-auth-dir

haxney:
`server-auth-dir' is used to store text files containing authentication keys. This is the only configuration file or directory of the package. This package does take care of creating the containing directory if necessary.


tarsius:
Long time no see. :wave:

> This is the only configuration file or directory of the package.

The docstring of `server-auth-dir` itself mentions `server-socket-dir`, so this is not true.

I think we shouldn't theme that second variable, because its default value is set dynamically to a very appropriate directory, which is also out of sight. But we should use a value for the former that would continue to work if we later changed our mind about that. I.e., use `var/server/auth/` so we could put `var/server/socket/` next to it.

haxney:
> Long time no see. 👋

Hey! I'm surprised you remember me! :heart:

> > This is the only configuration file or directory of the package.
> 
> The docstring of `server-auth-dir` itself mentions `server-socket-dir`, so this is not true.

It's only *kind of* not true. As you mention, `server-socket-dir` needs to be set to something that `emacsclient` can find without having to load the init file, so it isn't really something we can change in a package (which runs because it is called from the init file). I can note this in the PR and commit message. Should I overwrite and squash the original commit to update this or just create a second commit?

> I think we shouldn't theme that second variable, because its default value is set dynamically to a very appropriate directory, which is also out of sight. But we should use a value for the former that would continue to work if we later changed our mind about that. I.e., use `var/server/auth/` so we could put `var/server/socket/` next to it.

Yeah, I was on the fence about whether to do that. One extra level of nesting won't hurt, right? I'll go ahead and change it now.

PR #269, opened 2026-04-06
fsharp-ts-mode: Theme variables

dalugm:
Theme `fsharp-ts-eglot-server-install-dir' and
`fsharp-ts-repl-history-file'


tarsius:
Thanks!

PR #270, opened 2026-04-15
shell-maker: support new `shell-maker-root-path` variable

caisah:


caisah:
Latest version of `shell-maker` uses the `shell-maker-root-path` variable.  
For backwards compatibility I think is best to keep both 🤷🏻‍♂️
https://github.com/xenodium/shell-maker/blob/main/shell-maker.el#L99

tarsius:
Thanks!

